PR #18, opened 2024-04-09
LIB-297 fix(badge): font style and alignment

Trebh:
Jira: [[LIB-297]](https://fiscozen.atlassian.net/browse/LIB-297)


packages/badge/src/FzBadge.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div :class="classes">
+  <div :class="classes" class="h-20 flex items-center font-medium">
```
oxcened, line 2: there are already fixed classes in `classes`, I would keep them all in one place, either here or in `classes`

oxcened (approved):

PR #32, opened 2024-05-07
LIB-391 docs(topbar): add RFC

oxcened:
Jira: [LIB-391](https://fiscozen.atlassian.net/browse/LIB-391)

### DIFF
--- a/docs/RFC/topbar/2024-05-07.md
+++ b/docs/RFC/topbar/2024-05-07.md
@@ -1,0 +1,50 @@
+# RFC Template
+
+**Feature Name:** topbar
+
+**Start Date:** 2024-05-07
+
+**Author:** Alen Ajam
+
+**Related components/issues:** button, icon-button, link
+
+## Summary
+
+Use the topbar when you want to display an important message, and provides actions for users to address.
+
+## Motivation
+
+The main application needs a topbar component. It should encapsulate the UI logic expressed in the following [Figma component](https://www.figma.com/file/tqDILjLuuGCXICMbLrzxB4/Design-System?type=design&node-id=838-28878&mode=design&t=ZcwABejnK4pajebs-0).
+
+## Detailed design
+
+A Topbar component that is intended to be placed at the top of a page.
+
+The text is provided through the `text` slot.
+The action component is provided through slots, the component then internally renders the desired slot based on the style chosen and the current viewport.
+
+### Topbar props
+
+| Name  | Type                                                      | Required | Default   |
+|-------|-----------------------------------------------------------|----------|-----------|
+| type  | "default" \| "danger"                                     | No       | "default" |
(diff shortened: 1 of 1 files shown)


docs/RFC/topbar/2024-05-07.md
```diff
@@ -0,0 +1,50 @@
...
+
+**Related components/issues:** button, icon-button, link
+
+## Summary
+
+Use the topbar when you want to display an important message, and provides actions for users to address.
+
+## Motivation
+
+Why are we doing this? What use cases does it support? What is the expected outcome?
```
Trebh, line 17: Either we put the answer to this question or we remove the section, this is a prompt :laughing:

oxcened: Must've missed it 🤣

oxcened: fixed 👍

docs/RFC/topbar/2024-05-07.md
```diff
@@ -0,0 +1,50 @@
...
+| Name        | Required | Notes                                             |
+|-------------|----------|---------------------------------------------------|
+| text        | Yes      |                                                   |
+| button      | No       | Applies if type is "button" or "hybrid"           |
+| icon-button | No       | Applies if type prop is "icon-button" or "hybrid" |
+| link        | No       | Applies if type prop is "link"                    |
+
+## Drawbacks
+
+Providing the entire action component (button/icon-button/link) as a slot can be verbose, but it enables the user to completely customize it.
```
Crish15, line 44: I have some doubts about the use of all these slots. How do you imagine the usage by a user? For example, from what I understand I should write the following code to display a button in the action section:

```
<FzTopBar style='button'>
   <template #text>
   very very long text
   </template>
   <template #button>
       <FzButton ... />
   </template>
</FzTopBar>
```

right?

Crish15: Okay then you could simplify it to a slot named “action” where you would insert the button, or link, or other components.
For hybrid mode handling you could provide an additional “action-hybrid” slot (I couldn't think of any other name ahah) where you could insert the component to be displayed in mobile device alternating with the normal action slot.
With this approach we also get rid of dependencies on other components. Obviously the component becomes less tied to the design system, so I don't know if that is really what we want

ex:
```
<FzTopBar hybrid=true>
   very very long text
   <template #action>
       <FzButton ... />
   </template>
   <template #action-mobile>
     <FzList .../>
  </template>
</FzTopBar>

```
I would use the default slot for text: the main purpose of the TopBar is to display text, so I think it is not necessary to specify a specific slot or props.

oxcened: Yeah that looks better. The hybrid prop could be redundant as we could just check if both the action and action-mobile slots exist.

Crish15 (approved):

Trebh (approved):

PR #44, opened 2024-05-21
feat(LIB-392): init topbar component

oxcened:
Jira: [LIB-392](https://fiscozen.atlassian.net/browse/LIB-392)

### DIFF
--- a/packages/topbar/src/FzTopbar.vue
+++ b/packages/topbar/src/FzTopbar.vue
@@ -1,6 +1,8 @@
 <template>
   <div :class="containerClass">
-    <span> Testo qui </span>
+    <span class="text-xs lg:text-sm">
+      <slot></slot>
+    </span>
 
     <slot name="action">
       <FzButton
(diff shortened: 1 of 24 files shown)


Trebh (approved):
lgtm

packages/topbar/src/FzTopbar.vue
```diff
@@ -0,0 +1,125 @@
+<template>
+  <div :class="containerClass">
+    <span> Testo qui </span>
```
Crish15, line 3: You should probably use `<slot></slot>` here and not this static text

Crish15 (changes requested):

packages/topbar/src/FzTopbar.vue
oxcened: oh yeah my bad, good catch

oxcened: done in 13063ef

Crish15 (approved):

PR #51, opened 2024-06-04
feat(LIB-468): FzCard

Crish15:
Jira: [LIB-468](https://fiscozen.atlassian.net/browse/LIB-468)


packages/card/src/FzCard.vue
```diff
@@ -0,0 +1,87 @@
...
+    <header :class="[headerStaticClass, borderColor]">
+      <h2 class="text-core-black text-2xl m-0 p-0">{{ props.title }}</h2>
+    </header>
+    <article :class="['p-20 flex-1', contentClass]">
+      <slot></slot>
+    </article>
+    <footer v-if="atLeastOneButton" :class="[footerStaticClass, borderColor]">
+      <FzIconButton
+        v-if="tertiaryAction"
+        @click="tertiaryAction.callback"
```
Crish15, line 12: emit

packages/card/src/types.ts
```diff
@@ -0,0 +1,22 @@
+export type FzCardProps = {
+  title: string;
+  color?: FzCardColor;
```
oxcened, line 3: I'd advise against typing props as enum, as the developer is then forced to import the enum. Better off using a union type.

packages/card/src/types.ts
```diff
@@ -0,0 +1,22 @@
...
+  primaryAction?: FzCardButton;
+  secondaryAction?: FzCardButton;
+  tertiaryAction?: FzCardIconButton;
```
oxcened, line 6: What's the benefit of taking objects as props? I think primitive ones are easier to pass

packages/card/src/types.ts
Crish15: I implemented the enum so that the end user is forced to use only the value defined within the enum. We should probably discuss this internally and choose what is best for us. For now, I will switch to the union type

packages/card/src/types.ts
Crish15: I left the object in case you want to extend it with other properties in the future

packages/card/src/types.ts
oxcened: The union type lets you achieve the same without the need to import anything.
This was actually already discussed very early in the design system: https://github.com/fiscozen/design_system/pull/3#discussion_r1515771633

Trebh (approved):

PR #55, opened 2024-06-05
fix(LIB-478): tabs delete auto switch to existing tab

Crish15:
Jira: [LIB-478](https://fiscozen.atlassian.net/browse/LIB-478)

### DIFF
--- a/packages/tab/src/FzTab.vue
+++ b/packages/tab/src/FzTab.vue
@@ -3,10 +3,11 @@
 </template>
 
 <script setup lang="ts">
-import { inject, onMounted, Ref, computed } from "vue";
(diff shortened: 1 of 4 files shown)


packages/tab/src/FzTab.vue
```diff
@@ -3,10 +3,11 @@
 </template>
 
 <script setup lang="ts">
-import { inject, onMounted, Ref, computed } from "vue";
+import { inject, onMounted, Ref, computed, onBeforeUnmount } from "vue";
 import { FzTabProps } from "./types";
 
 const props = defineProps<FzTabProps>();
+const emit = defineEmits(["unmount"]);
```
Crish15, line 10: rimuovere

PR #59, opened 2024-06-10
feat(LIB-475): select

Crish15:
Jira: [LIB-475](https://fiscozen.atlassian.net/browse/LIB-475)


packages/select/src/index.ts
```diff
@@ -0,0 +1 @@
+export { default as FzSelect } from "./FzSelect.vue";
```
oxcened, line 1: I'd also export the types

packages/select/src/components/FzSelectOption.vue
```diff
@@ -0,0 +1,32 @@
...
+  selectedValue: string;
+}>();
+
+const emit = defineEmits(["click"]);
+const staticClass =
+  "flex items-center text-left h-40 bg-white text-blue-500 font-medium cursor-pointer py-6 px-12 rounded text-md";
+const computedClass = computed(() => [
+  props.selectedValue === props.option.value
+    ? "!bg-background-alice-blue"
+    : "hover:!bg-background-alice-blue !text-core-black hover:!text-blue-500",
```
oxcened, line 30: why's everything important?

packages/select/src/components/FzSelectOption.vue
```diff
@@ -0,0 +1,32 @@
+<template>
+  <button
+    @click="$emit('click')"
```
oxcened, line 3:
```suggestion
    @click="emit('click')"
```
this way you're using the return value of `defineEmits`

packages/select/src/FzSelect.vue
```diff
@@ -0,0 +1,158 @@
...
+  return props.options.find((option) => option.value === model.value);
+});
+
+watch(() => [props.size, model.value], calculateContainerWidth);
+
+onMounted(() => {
+  calculateContainerWidth();
+});
+
+const onSelect = (value: string) => {
```
oxcened, line 130:
```suggestion
const handleSelect = (value: string) => {
```

packages/select/src/FzSelect.vue
```diff
@@ -0,0 +1,158 @@
...
+  model.value = value;
+  emit("select", value);
+  isOpen.value = false;
+};
+
+const onPickerClick = () => {
```
oxcened, line 136:
```suggestion
const handlePickerClick = () => {
```

packages/select/src/FzSelect.vue
```diff
@@ -0,0 +1,158 @@
...
+
+const staticSpanClass =
+  "overflow-hidden text-ellipsis whitespace-nowrap flex-[1]";
+const computedSpanClass = computed(() => [
+  selectedOption.value && !props.disabled ? "text-core-black font-medium" : "text-grey-300",
+]);
+
+const computedHelpClass = computed(() => [
+  props.size === "sm" ? "text-xs" : "",
+  props.size === "md" ? "text-sm" : "",
```
oxcened, line 110:
```suggestion
  props.size === "sm" ? "text-xs" : "text-sm",
```

packages/select/src/FzSelect.vue
```diff
@@ -0,0 +1,158 @@
...
+
+const emit = defineEmits(["select"]);
+
+const staticPickerClass =
+  "flex justify-between items-center px-10 border-1 w-full rounded gap-8 text-left";
+const computedPickerClass = computed(() => [
+  `text-${props.size}`,
+  props.size === "sm" ? "h-24" : "",
+  props.size === "md" ? "h-32" : "",
+  props.size === "lg" ? "h-40" : "",
```
oxcened, line 93: you can nest an object inside of the array, would work better for this use case

packages/select/src/FzSelect.vue
```diff
@@ -0,0 +1,158 @@
...
+const isOpen = ref(false);
+const opener = ref<HTMLElement>();
+const containerWidth = ref<string>("auto");
+
+const emit = defineEmits(["select"]);
+
+const staticPickerClass =
+  "flex justify-between items-center px-10 border-1 w-full rounded gap-8 text-left";
+const computedPickerClass = computed(() => [
+  `text-${props.size}`,
```
oxcened, line 90: you can't dynamically create tailwind classes

Trebh:
Very risky question (that needs to be answered though, I can guarantee it will be asked): couldn't we wrap the native <select> and   add style/functionality to that?

oxcened:
> Very risky question (that needs to be answered though, I can guarantee it will be asked): couldn't we wrap the native `<select>` and add style/functionality to that?

Much better solution accessibility-wise, but we wouldn't be able to customize it as per design requirements.

Crish15:
> Very risky question (that needs to be answered though, I can guarantee it will be asked): couldn't we wrap the native `<select>` and add style/functionality to that?

I took inspiration from other library and everyone creates his own custom component. Currently with buttons and title change based on the selected value, accessibility should be ensured. Otherwise I could change the main button to a text type input

packages/select/src/FzSelect.vue
Crish15: following conventions, having separate ifs helps to better understand the different cases. Also, in case the size "lg" or "xs" is added,  it is easier to add the new feature

Trebh (approved):
possible improvement: avoid the necessity to map options objects at the application level by making the `value` and `label` keys dynamic?

PR #67, opened 2024-06-26
fix id and label for not unique

Crish15:
Some fix I made to make the radio work for ticket [LIB-491](https://fiscozen.atlassian.net/browse/LIB-491)

### DIFF
--- a/packages/radio/src/FzRadio.vue
+++ b/packages/radio/src/FzRadio.vue
@@ -2,7 +2,7 @@
   <div class="flex justify-center flex-col w-fit">
     <input
       type="radio"
-      :id="label"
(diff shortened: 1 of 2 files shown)


packages/radio/src/FzRadio.vue
```diff
@@ -2,7 +2,7 @@
   <div class="flex justify-center flex-col w-fit">
     <input
       type="radio"
-      :id="label"
+      :id="name ? name + '-' + label : label"
```
Trebh, line 5: I'd always choose template strings over `+` operator on strings

Trebh (approved):

PR #71, opened 2024-07-04
LIB-517 feat(alert): add link as cta

oxcened:
Jira: [LIB-517](https://fiscozen.atlassian.net/browse/LIB-517)

### DIFF
--- a/packages/alert/src/types.ts
+++ b/packages/alert/src/types.ts
@@ -1,3 +1,5 @@
+import { RouteLocationRaw } from 'vue-router'
+
 type AlertProps = {
   /**
    * Type of the topbar which dictates the appearance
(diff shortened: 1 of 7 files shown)


packages/alert/src/types.ts
```diff
@@ -23,10 +41,6 @@ type AlertProps = {
    * Title of the alert (applies only if style is not simple)
    */
   title?: string
-  /**
-   * Hide the action button (applies only if style is not simple)
-   */
-  hideAction?: boolean
```
Crish15, line 29: I need this property for the alert inside Dichiarazioni (it's a collapsable alert without an explicit action or link)

Crish15 (changes requested):

Crish15 (approved):

PR #74, opened 2024-07-10
fix(LIB-500): checkbox fix for main app

Crish15:
Jira: [LIB-500](https://fiscozen.atlassian.net/browse/LIB-500)


packages/checkbox/src/FzCheckboxGroup.vue
```diff
@@ -26,11 +26,15 @@
...
-import { FzCheckboxGroupProps, ParentCheckbox } from "./types";
+import { FzCheckboxGroupProps } from "./types";
 import FzCheckboxErrorText from "./components/FzCheckboxErrorText.vue";
 import { mapSizeToClasses } from "./common";
 import FzCheckboxGroupOption from "./components/FzCheckboxGroupOption.vue";
 
+FzCheckboxGroupOption.compatConfig = {
+  MODE:3
+}
+
```
oxcened, line 37: can't this be put in the app and not here?

packages/checkbox/src/components/FzCheckboxGroupOption.vue
```diff
@@ -32,9 +32,13 @@ import { computed } from "vue";
 import FzCheckbox from "../FzCheckbox.vue";
 import { ParentCheckbox } from "../types";
 
-const props = defineProps<ParentCheckbox & { size: "sm" | "md" }>();
+FzCheckbox.compatConfig = {
+  MODE: 3,
+};
```
oxcened, line 37: same as above

Crish15: To avoid potential problems, I have included it here as well, because I am not sure that if you set compatConifg for FzCheckbox in the production application on a page, it will address all FzCheckboxes within that page (even nested ones like this one).

packages/checkbox/src/FzCheckboxGroup.vue
Crish15: No because FzCheckboxGroupOption is not exposed, therefor I cannot set the mode:3 in the app

Trebh (approved):

PR #79, opened 2024-07-16
Feat/lib 555 datepicker component

Trebh:
[Jira](https://fiscozen.atlassian.net/browse/LIB-555)

### DIFF
--- a/packages/datepicker/package.json
+++ b/packages/datepicker/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@fiscozen/datepicker",
-  "version": "0.1.0",
+  "version": "0.1.1",
   "description": "Design System Datepicker component",
   "scripts": {
     "coverage": "vitest run --coverage",
--- a/packages/datepicker/package.json
+++ b/packages/datepicker/package.json
@@ -36,5 +36,8 @@
     "vitest": "^1.2.0",
     "vue-tsc": "^1.8.25"
   },
-  "license": "MIT"
+  "license": "MIT",
+  "dependencies": {
+    "@vuepic/vue-datepicker": "^8.8.1"
+  }
 }
--- a/packages/datepicker/package.json
+++ b/packages/datepicker/package.json
@@ -38,6 +38,10 @@
   },
   "license": "MIT",
   "dependencies": {
-    "@vuepic/vue-datepicker": "^8.8.1"
+    "@fiscozen/composables": "workspace:^",
+    "@fiscozen/icons": "^0.1.4",
+    "@fiscozen/style": "workspace:^",
+    "@vuepic/vue-datepicker": "^8.8.1",
+    "date-fns": "^3.6.0"
   }
 }
--- a/packages/datepicker/package.json
+++ b/packages/datepicker/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@fiscozen/datepicker",
-  "version": "0.1.3",
+  "version": "0.1.4",
   "description": "Design System Datepicker component",
   "scripts": {
     "coverage": "vitest run --coverage",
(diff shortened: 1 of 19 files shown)


packages/datepicker/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "typescript": "~5.3.0",
+    "vite": "^5.0.10",
+    "vite-plugin-dts": "^3.8.3",
+    "vitest": "^1.2.0",
+    "vue-tsc": "^1.8.25"
+  },
+  "license": "MIT",
+  "dependencies": {
+    "@fiscozen/composables": "workspace:^",
+    "@fiscozen/icons": "^0.1.4",
```
oxcened, line 42: did you fix the version on purpose?

packages/datepicker/src/FzDatepicker.vue
```diff
@@ -0,0 +1,163 @@
...
+    box-shadow: 0px 0px 0px 0px #000;
+}
+
+.dp__month_year_wrap .dp__inner_nav:hover {
+    color: var(--core-white, #FFF);
+}
+
+.dp__month_year_wrap {
+    height: 32px;
+}
```
oxcened, line 162: we COULD use @apply in this case, would be nice to keep using tailwind classes even here

packages/datepicker/src/__tests__/FzDatepicker.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+  it("matches snaphost", async ({ expect }) => {
+    const wrapper = mount(FzDatepicker, {
+      props: {
+        modelValue: new Date()
+      },
+      slots: {},
+    });
+
+    expect(wrapper.html()).toMatchSnapshot();
+  });
```
oxcened, line 20: I saw there's a bunch of different stories, you could do a snapshot test for each

packages/datepicker/package.json
Trebh: I did, but maybe not the best idea since it already screwed up chromatic builds

packages/datepicker/src/FzDatepicker.vue
Trebh: in this case I would keep it this way since it is a brutal override that I don't really like, but we could change it to apply, yes

Crish15:
I was trying to use the `0.1.1` version, but I am having some problems:

- The datepicker has a reddish background.
- When selecting a value, the label goes over the icon (seems to be missing some padding on the left) 
image

- When used inside a FzConfirmDialog, the DatePicker dropdown creates an overflow. I would expect the container to always stay on top without going to affect the container
image

Nice to have: 
Add a label to be passed via props, as other components do (such as FzInput, FzTextarea, etc.).

Trebh:
> I was trying to use the `0.1.1` version, but I am having some problems:
> 
> * The datepicker has a reddish background.
> * When selecting a value, the label goes over the icon (seems to be missing some padding on the left)
>   image

I will look into this: it seems that the input date is "wrong", maybe a formatting issue

> * When used inside a FzConfirmDialog, the DatePicker dropdown creates an overflow. I would expect the container to always stay on top without going to affect the container
>   image

This is an issue that can't be solved entirely in the Datepicker component. I can add move the datepicker to fixed position, but we also need to merge 
https://github.com/fiscozen/design_system/pull/106
in order for that to work

> Nice to have: Add a label to be passed via props, as other components do (such as FzInput, FzTextarea, etc.).
This will not be part of the first implementation as agreed in previous meetings: the datepicker will be using our FzInput therefore we don't want to reimplement things like label etc

Crish15 (approved):

PR #81, opened 2024-07-16
Fix(LIB-491): design radio 

Crish15:
fix for Jira: [(LIB-491)](https://fiscozen.atlassian.net/browse/(LIB-491))

### DIFF
--- a/packages/radio/src/FzRadio.vue
+++ b/packages/radio/src/FzRadio.vue
@@ -39,7 +39,8 @@ const computedId = computed(() => props.name ? `${props.name}-${props.label}` : 
 
 const radioContainer = ref<HTMLInputElement | null>(null);
 const staticLabelClass = `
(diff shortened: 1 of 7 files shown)


packages/radio/src/FzRadio.vue
```diff
@@ -65,7 +66,7 @@ const computedLabelClass = computed(() => [
     ? "before:h-12 before:w-12 before:mt-[3px] peer-checked:before:border-4"
     : "",
   props.size === "md"
-    ? "before:h-16 before:w-16 before:mt-4 peer-checked:before:border-[5px]"
+    ? "before:h-16 before:w-16 peer-checked:before:border-[5px]"
     : "",
```
oxcened, line 70: an object would me more suited for toggling classes dynamically

oxcened (approved):

PR #82, opened 2024-07-16
fix(LIB-523): FzCard fixes for production app

Crish15:
Jira: [LIB-523](https://fiscozen.atlassian.net/browse/LIB-523)

### DIFF
--- a/packages/card/src/types.ts
+++ b/packages/card/src/types.ts
@@ -1,12 +1,39 @@
 export type FzCardProps = {
+  /**
+   * The title of the card
+   */
   title: string;
+  /**
+   * The background color of the card
+   */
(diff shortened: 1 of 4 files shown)


packages/card/src/types.ts
```diff
@@ -7,6 +7,7 @@ export type FzCardProps = {
   contentClass?: string;
   collapsible?: boolean;
   defaultExpanded?: boolean;
+  alwaysAlive?: boolean;
```
oxcened, line 10: All props are missing documentation which we usually write through /** comments. It isn't clear what alwaysAlive in particular does. I'd add those and make sure that storybook correctly displays them.

oxcened (approved):

PR #86, opened 2024-07-22
fix(LIB-571): button text ellipsis overflow

Crish15:
Jira: [LIB-571](https://fiscozen.atlassian.net/browse/LIB-571)

### DIFF
--- a/packages/button/src/FzButton.vue
+++ b/packages/button/src/FzButton.vue
@@ -9,9 +9,11 @@
         />
       </slot>
     </div>
-    <slot>
-      {{ label }}
-    </slot>
+    <div :class="['truncate', containerClass]">
+      <slot>
+        {{ label }}
+      </slot>
+    </div>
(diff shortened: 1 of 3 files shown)


packages/button/src/FzButton.vue
```diff
@@ -9,9 +9,11 @@
         />
       </slot>
     </div>
-    <slot>
-      {{ label }}
-    </slot>
+    <div :class="['truncate', containerClass]">
```
oxcened, line 12: For now it's all right, I would later on define a standard for all class props across all components. For example, in this case I wouldn't be able to get rid of the truncate class if I wanted to, an idea could be an override prop.

Crish15: In my opinion truncate should not be overridden because it is defined as a rule in the design system. In other cases I agree with you and we should study how to override some props/style.

oxcened (approved):

PR #88, opened 2024-07-23
feat(LIB-541): FzInput

Crish15:
Jira: [LIB-541](https://fiscozen.atlassian.net/browse/LIB-541)


packages/input/src/FzInput.vue
```diff
@@ -0,0 +1,97 @@
+<template>
+  <div class="w-full flex flex-col gap-8" @click="inputRef?.focus()">
+    <label :class="['text-sm', computedLabelClass]">
```
oxcened, line 3: label is not binded to input, see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input

oxcened (approved):

PR #103, opened 2024-08-01
LIB-636 feat(toast): open stack when new toast is enqueued

oxcened:
Jira: [LIB-636](https://fiscozen.atlassian.net/browse/LIB-636)

### DIFF
--- a/packages/toast/src/FzToastQueue.vue
+++ b/packages/toast/src/FzToastQueue.vue
@@ -1,6 +1,7 @@
 <template>
   <div
     class="hoverable group"
+    :class="{ hovering: isHovering }"
     @mouseenter="isHovering = true"
     @mouseleave="isHovering = false"
   >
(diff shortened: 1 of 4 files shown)


packages/toast/src/FzToastQueue.vue
```diff
@@ -37,16 +42,29 @@
...
+const toastsLength = computed(() => toasts.value.length);
+let currentTimeoutId: number | undefined = undefined;
+
+watch(toastsLength, (newValue, oldValue) => {
+  if (!props.openOnNewToast) return;
+  if (newValue <= oldValue) return;
+  if (currentTimeoutId != null) window.clearTimeout(currentTimeoutId);
+  
+  isHovering.value = true;
+  currentTimeoutId = window.setTimeout(() => (isHovering.value = false), 2000);
```
Crish15, line 66: I suggest adding the ability to customize the maximum timeout duration, perhaps via a custom props ( with the default value of 2000)

oxcened: fair enough

Crish15 (approved):

PR #105, opened 2024-08-02
LIB-641 feat(dialog): escape key listener

Trebh:
[Jira](https://fiscozen.atlassian.net/browse/LIB-641)


Trebh:
I also changed the name of the event, I don't see counterpoints

oxcened (commented):
you should reuse this one
https://github.com/fiscozen/design_system/blob/main/packages/composables/src/composables/useKeyDown.ts

Trebh:
> you should reuse this one https://github.com/fiscozen/design_system/blob/main/packages/composables/src/composables/useKeyDown.ts

I might add a keyUp then? Usually I prefer that over keyDown to avoid multiple calls if the user keeps pressing longer than needed

oxcened:
> > you should reuse this one https://github.com/fiscozen/design_system/blob/main/packages/composables/src/composables/useKeyDown.ts
> 
> I might add a keyUp then? Usually I prefer that over keyDown to avoid multiple calls if the user keeps pressing longer than needed

yeah sure

oxcened (approved):

PR #126, opened 2024-08-26
LIB-797 feat(input): expose inputRef

oxcened:
Jira: [LIB-797](https://fiscozen.atlassian.net/browse/LIB-797)

### DIFF
--- a/packages/input/src/FzCurrencyInput.vue
+++ b/packages/input/src/FzCurrencyInput.vue
@@ -3,24 +3,26 @@
 </template>
 
 <script setup lang="ts">
-import { onMounted, ref } from "vue";
+import { computed, onMounted, ref } from "vue";
 import FzInput from "./FzInput.vue";
 import { FzCurrencyInputProps } from "./types";
 import { useCurrency } from "@fiscozen/composables";
 
-const fzInputRef = ref();
+const fzInputRef = ref<InstanceType<typeof FzInput>>();
+const containerRef = computed(() => fzInputRef.value?.containerRef);
+const inputRef = computed(() => fzInputRef.value?.inputRef);
 const props = defineProps<FzCurrencyInputProps>();
-const { inputRef } = useCurrency();
+const { inputRef: currencyInputRef } = useCurrency();
 
 defineEmits(["update:amount"]);
 
 onMounted(() => {
-  inputRef.value = fzInputRef.value.inputRef;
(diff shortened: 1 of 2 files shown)


packages/input/src/FzCurrencyInput.vue
```diff
@@ -18,4 +18,9 @@ onMounted(() => {
   inputRef.value = fzInputRef.value.inputRef;
 });
 const model = defineModel("amount");
+
+defineExpose({
+  inputRef: fzInputRef.value?.inputRef,
```
Crish15, line 23: are you sure this will be the inputRef and not null/undefined? I think defineExpose is executed before fzInputRef is populated so the value will be null

oxcened: yes I'm looking at it right now, indeed doesn't work

Trebh: why not exposing the ref instead of the value?

packages/input/src/FzCurrencyInput.vue
```diff
@@ -3,19 +3,26 @@
...
+const inputRef = computed(() => fzInputRef.value?.inputRef);
 const props = defineProps<FzCurrencyInputProps>();
-const { inputRef } = useCurrency();
+const { inputRef: currencyInputRef } = useCurrency();
 
 defineEmits(["update:amount"]);
 
 onMounted(() => {
-  inputRef.value = fzInputRef.value.inputRef;
+  currencyInputRef.value = inputRef.value;
```
Trebh, line 20: are we sure we don't need to watch for inputRef changes?

oxcened: what for?

Trebh: I'm not sure what happens if the user for example puts the `FzCurrencyInput` in a v-if block and toggles it, or if it is inside a v-for with index based keys and makes modification to the list... and so on.
This runs once, are we sure that we don't want the ref but the value?

oxcened: why should this be an issue related to my changes though?

oxcened: in both cases we are assigning to the input's ref on the component's mount, it's one-off read, so nothing really changed

Trebh (approved):

Crish15 (approved):

PR #128, opened 2024-08-28
FzTypeahead filteredOptions deprecated and debounce fix

Crish15:
~~Motivation:~~

~~I believe that handling the private state of the FzTypeahead, specifically the 'filteredOptions,' is not the most appropriate approach. Instead, I have created a callback function that returns a promise of FzSelectOptions, which is invoked (with a debounce) each time a user types into the input. This allows the end user to avoid creating a ref within the components for the filtered options and simply link the API call (or a function that transforms the API call to the correct format) to the FzTypeahead.~~

Deprecated `filteredOptions` in favor of using only `options`. Added `filterable` prop to enable or disable the ability to filter options by typing in the input text (enabled by default).

Additionally, I have fixed the debounce issue: previously, it was called every time with a different timer id, leading to a queue of functions being created and launched. Now, the correct function should be launched just once.

### DIFF
--- a/apps/storybook/src/stories/Typeahead.stories.ts
+++ b/apps/storybook/src/stories/Typeahead.stories.ts
@@ -194,6 +194,48 @@ const RemoteLoading: Story = {
   }
 }
 
-export { Default, Precompiled, PrecompiledObject, NoDelayTime, HundredOptions, RemoteLoading }
+async function remoteCallback(text?: string) {
+  const res = await fetch(`https://dummyjson.com/users/search?q=${text}`)
+  const data = await res.json()
+
+  return data.users.map((user: any) => ({
+    label: user.firstName + ' ' + user.lastName,
+    value: user.id
+  }))
+}
+
+const RemoteLoadingWithRemoteFunction: Story = {
+  render: (args) => ({
+    components: {FzTypeahead},
+    setup() {
+      const text = ref();
+      return {
+        text,
+        args
+      }
+    },
+    methods: {
+    },
+    template: `
+      <div class="h-[100vh] w-[100-vw] p-16">
+        <FzTypeahead v-bind="args" v-model="text" @fztypeahead:input="onInputChange"/>
+      </div>
+    `
+  }),
+  args: {
+    inputProps: {
+      label: 'This is a label',
+      placeholder: 'This is a placeholder'
+    },
+    selectProps: {
+      isOpen: false,
+      options: []
+    },
+    remoteFn: remoteCallback
+  }
+
+}
+
+export { Default, Precompiled, PrecompiledObject, NoDelayTime, HundredOptions, RemoteLoading, RemoteLoadingWithRemoteFunction }
 
 export default meta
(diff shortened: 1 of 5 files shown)


Trebh:
> Motivation:
> 
> I believe that handling the private state of the FzTypeahead, specifically the 'filteredOptions,' is not the most appropriate approach. Instead, I have created a callback function that returns a promise of FzSelectOptions, which is invoked (with a debounce) each time a user types into the input. This allows the end user to avoid creating a ref within the components for the filtered options and simply link the API call (or a function that transforms the API call to the correct format) to the FzTypeahead.
> 
> Additionally, I have fixed the debounce issue: previously, it was called every time with a different timer id, leading to a queue of functions being created and launched. Now, the correct function should be launched just once.

These are two separate issues, I would try to address them in separate PRs.

I _really_ dislike the callback approach to be honest. While it's arguably more "tidy" since we hide the ref of the results from the user, it still couples (albeit loosely) to a specific implementation, in this case an async Promise.
What if the caller wants to use websockets to get their options?
Generally speaking presentational components should not own data, ever, imho unless its a purely internal state. I would argue that results from a search coming frrom APIs do not make into this cathegory.

Since I can understand where this is coming from, could we keep both and leaving the baseline "escape hatch" of having a `filteredOptions` provided by the user that trumps all other stuff, and then instead applying remoteFn if we want to?

Crish15:
> > Motivation:
> > I believe that handling the private state of the FzTypeahead, specifically the 'filteredOptions,' is not the most appropriate approach. Instead, I have created a callback function that returns a promise of FzSelectOptions, which is invoked (with a debounce) each time a user types into the input. This allows the end user to avoid creating a ref within the components for the filtered options and simply link the API call (or a function that transforms the API call to the correct format) to the FzTypeahead.
> > Additionally, I have fixed the debounce issue: previously, it was called every time with a different timer id, leading to a queue of functions being created and launched. Now, the correct function should be launched just once.
> 
> These are two separate issues, I would try to address them in separate PRs.
> 
> I _really_ dislike the callback approach to be honest. While it's arguably more "tidy" since we hide the ref of the results from the user, it still couples (albeit loosely) to a specific implementation, in this case an async Promise. What if the caller wants to use websockets to get their options? Generally speaking presentational components should not own data, ever, imho unless its a purely internal state. I would argue that results from a search coming frrom APIs do not make into this cathegory.
> 
> Since I can understand where this is coming from, could we keep both and leaving the baseline "escape hatch" of having a `filteredOptions` provided by the user that trumps all other stuff, and then instead applying remoteFn if we want to?

Okay, I understand the issue with using a function that returns a promise, and it may not be the best solution. Could we consider removing `filteredOptions` and using `selectProps.options` instead? When we call an API, we're essentially reassigning the options, so `filteredOptions` might not be necessary. Additionally, it seems more straightforward and easier to understand if the end user just creates a single list of options rather than maintaining both `options` and `filteredOptions`.

Trebh (approved):

PR #149, opened 2024-10-09
feat(datepicker): update model early upon calendar date selection

Crish15:


oxcened:
@Trebh manca il tag mi sembra

PR #165, opened 2024-10-22
Fix/lib 1266 typeahead fix

Trebh:
Jira: [LIB-1266](https://fiscozen.atlassian.net/browse/LIB-1266)

### DIFF
--- a/packages/composables/src/FzFloating.vue
+++ b/packages/composables/src/FzFloating.vue
@@ -12,7 +12,7 @@ const props = withDefaults(defineProps<FzFloatingProps>(), {
 const emits = defineEmits(['fzfloating:setPosition'])
 
 const content = ref<HTMLElement | null>(null)
-const opener = ref(null)
+const opener = ref<HTMLElement | null>(null)
 
 const slots = useSlots()
 
@@ -37,14 +37,14 @@ const useFloatingOpts: FzUseFloatingArgs = {
   }
 }
 
-const testOpts = toRefs(useFloatingOpts)
+const dynamicOpts = toRefs(useFloatingOpts)
 if (slots.opener) {
   useFloatingOpts.opener = {
     domRef: opener
   }
 }
 
-const floating = useFloating(testOpts)
+const floating = useFloating(dynamicOpts)
 
 const setPositionWhenOpen = () => {
   if (scheduledAnimationFrame) {
@@ -82,8 +82,8 @@ watch(
     if (!newVal) {
       return
     }
-    if (testOpts.opener && testOpts.opener.value) {
-      testOpts.opener.value = {
+    if (dynamicOpts.opener && dynamicOpts.opener.value) {
+      dynamicOpts.opener.value = {
         domRef: newVal
       }
     }
--- a/packages/composables/src/FzFloating.vue
+++ b/packages/composables/src/FzFloating.vue
@@ -1,5 +1,5 @@
 <script lang="ts" setup>
-import { ref, useSlots, watch, toRef, computed, onMounted, onBeforeUnmount } from 'vue'
+import { ref, useSlots, watch, toRef, computed, onBeforeUnmount, toRefs } from 'vue'
 import { useFloating } from './composables'
 import { FzFloatingProps, FzUseFloatingArgs } from './types'
 
@@ -11,15 +11,15 @@ const props = withDefaults(defineProps<FzFloatingProps>(), {
 
 const emits = defineEmits(['fzfloating:setPosition'])
 
-const opener = ref(null)
 const content = ref<HTMLElement | null>(null)
+const opener = ref(null)
 
 const slots = useSlots()
 
 let scheduledAnimationFrame = false
 
 const useFloatingOpts: FzUseFloatingArgs = {
-  position: computed(() => props.position),
+  position: props.position,
   element: {
     // @ts-ignore
     domRef: content
@@ -28,18 +28,23 @@ const useFloatingOpts: FzUseFloatingArgs = {
     // @ts-ignore
     domRef: toRef(props.container || document.body)
   },
-  useViewport: computed(() => props.useViewport),
+  opener: {
+    domRef: toRef(null)
+  },
+  useViewport: props.useViewport,
   callback(...args) {
     emits('fzfloating:setPosition', ...args)
   }
 }
+
+const testOpts = toRefs(useFloatingOpts)
 if (slots.opener) {
   useFloatingOpts.opener = {
     domRef: opener
   }
 }
 
-const floating = useFloating(useFloatingOpts)
+const floating = useFloating(testOpts)
 
 const setPositionWhenOpen = () => {
   if (scheduledAnimationFrame) {
@@ -69,6 +74,19 @@ watch(
     content.value.style.left = '0px'
     content.value.style.transform = 'none'
     floating.setPosition()
+  }
+)
+watch(
+  () => props.overrideOpener,
+  (newVal) => {
+    if (!newVal) {
+      return
+    }
+    if (testOpts.opener && testOpts.opener.value) {
+      testOpts.opener.value = {
+        domRef: newVal
+      }
+    }
   }
 )
 onBeforeUnmount(() => {
--- a/packages/composables/src/composables/useFloating.ts
+++ b/packages/composables/src/composables/useFloating.ts
@@ -1,9 +1,9 @@
 import { FzFloatingPosition, FzRect, FzUseFloatingArgs } from '../types'
-import { ref, reactive, onUnmounted, Ref, nextTick } from 'vue'
+import { ref, reactive, onUnmounted, Ref, nextTick, ToRefs } from 'vue'
 import { calcRealPos, getHighestAvailableSpacePos } from '../utils'
 
 export const useFloating = (
-  args: FzUseFloatingArgs
+  args: ToRefs<FzUseFloatingArgs>
 ): {
   float: FzRect
   rect: Ref<DOMRect | undefined>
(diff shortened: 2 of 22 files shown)


packages/select/src/FzSelect.vue
```diff
@@ -66,14 +67,27 @@
...
     >
-      <FzSelectOption
-        v-for="option in visibleOptions"
-        :key="option.value"
-        @click="() => handleSelect(option.value)"
-        :option="option"
-        :size="size"
-        :selectedValue="model"
-      />
+      <template v-if="visibleOptions.length">
```
Crish15, line 70: instead of doing a `v-if` here, I would assign the default option inside `updateVisibleOption` when no options are available

packages/select/src/components/FzSelectOption.vue
```diff
@@ -6,6 +6,9 @@
     :title="option.label"
     @click="
       () => {
+        if (option.disabled || option.readonly) {
```
Crish15, line 9: I think it is preferable to assign the `disabled` attribute

packages/typeahead/src/FzTypeahead.vue
```diff
@@ -1,17 +1,53 @@
 <template>
   <FzSelect
+    ref="fzselect"
     @select="(val) => emit('fztypeahead:select', val)"
+    :disabled
```
Crish15, line 5: disabled is not defined inside FzSelect

packages/select/src/FzSelect.vue
Trebh: I really don't like to mess with the state of the component... we are not assigning a "real" value here, it's just a visible feedback for the user

packages/select/src/components/FzSelectOption.vue
Trebh: do you mean to have just option.disabled here? I had to add readonly because there were 2 different visuals for those but the same behavior

packages/typeahead/src/FzTypeahead.vue
```diff
@@ -99,20 +161,23 @@ const debounceHandleInput = debounce(
...
 );
 
-function handleInput(val: string) {
+function handleInput(val: string, isOpen: boolean) {
   inputValue.value = val;
-  const selected = internalOptions.value.find((opt) => opt.value === val);
+  const selected = internalOptions.value?.find((opt) => opt.value === val);
   if (!selected) model.value = undefined;
   debounceHandleInput(val);
+  if (!isOpen) {
```
Crish15, line 169: are you sure you can use `isOpen` here? I think it is not possible, because it is not accessible outside the scope of FzSelect's slot. If you want two-way communication, you should probably use a v-model or provide/inject

packages/select/src/components/FzSelectOption.vue
Crish15: No, I mean that it is preferable to use the disabled attribute to disable the click.
eg:
`:title="option.label"
:disabled="option.disabled || option.readonly"`

packages/typeahead/src/FzTypeahead.vue
```diff
@@ -99,20 +161,23 @@ const debounceHandleInput = debounce(
...
+    fzselect.value.forceOpen();
+  }
 }
 
-const internalOptions = computed<FzSelectOptionsProps[]>(() => {
-  let res = props.selectProps.options;
+const internalOptions = computed<FzSelectOptionsProps[] | undefined>(() => {
```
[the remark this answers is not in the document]
Trebh: it's already "caught" in `safeSelectOpts` and given [] as default

packages/typeahead/src/FzTypeahead.vue
Crish15: Yeah, you are right... I read too much code, sorry 😜

packages/typeahead/src/FzTypeahead.vue
Trebh: sure you can, I pass isOpen to the function called inside the named slot. I don't want two way communication: i am just reading its state and calling FzSelect's `forceOpen` method if needed be

packages/typeahead/src/FzTypeahead.vue
```diff
@@ -1,17 +1,53 @@
...
-        ref="opener"
-        v-bind="inputProps"
-        :modelValue="inputValue"
-        @update:modelValue="handleInput"
-        @focus="() => !isOpen && handlePickerClick()"
-      ></FzInput>
+      <div class="w-full flex gap-8" ref="openerContainer">
+        <FzInput
+          ref="opener"
+          v-bind="inputProps"
```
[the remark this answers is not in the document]
Trebh: inputProps should win only where we don't want to have a final say. And we want to have a final say in label and placeholder usage, of that I would be pretty sure.

As per the computedProp: i generally agree, but it turned out to be a lot more of a hassle to do in practice and I opted out after trying

packages/typeahead/src/FzTypeahead.vue
Crish15: I thought it was "global", i missed the function call, my bad

packages/typeahead/src/FzTypeahead.vue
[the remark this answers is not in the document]
Crish15: As I understand it, the purpose of inputProps is to be used when you want to go into ‘expert mode’ and have control over everything. Maybe let's also hear from others to decide what the right strategy is.

packages/select/src/FzSelect.vue
Crish15: In any case, the model will have `‘’ ` as its value when there are no options. In the end you have created a ‘button’ with the value `‘’` anyway, so I think it is also more readable to have a single v-for on all options and in the function that populates these options handle the correct case

packages/typeahead/src/FzTypeahead.vue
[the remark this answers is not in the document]
Trebh: actually, I tried again, and you are just right. I will update the implementation

Crish15 (approved):

packages/composables/src/FzFloating.vue
```diff
@@ -11,15 +11,15 @@ const props = withDefaults(defineProps<FzFloatingProps>(), {
 
 const emits = defineEmits(['fzfloating:setPosition'])
 
-const opener = ref(null)
 const content = ref<HTMLElement | null>(null)
+const opener = ref(null)
```
oxcened, line 15: typing is missing

packages/composables/src/FzFloating.vue
```diff
@@ -28,18 +28,23 @@ const useFloatingOpts: FzUseFloatingArgs = {
...
+  opener: {
+    domRef: toRef(null)
+  },
+  useViewport: props.useViewport,
   callback(...args) {
     emits('fzfloating:setPosition', ...args)
   }
 }
+
+const testOpts = toRefs(useFloatingOpts)
```
oxcened, line 40: what is testOpts? the name is not really clear

packages/composables/src/composables/useFloating.ts
```diff
@@ -82,7 +82,13 @@ export const useFloating = (
       let translateY = 0
       let translateX = 0
 
-      if (args.opener && safeOpenerDomRef.value && openerRect && openerRect.value) {
+      if (
+        args.opener &&
+        args.opener.value &&
+        safeOpenerDomRef.value &&
+        openerRect &&
+        openerRect.value
```
oxcened, line 90:
```suggestion
        args.opener?.value &&
        safeOpenerDomRef.value &&
        openerRect?.value
```

packages/composables/src/composables/useFloating.ts
```diff
@@ -304,7 +310,9 @@ export const useFloating = (
 
       rect.value = safeElementDomRef.value.getBoundingClientRect()
 
-      args.callback && args.callback(rect, openerRect, containerRect, position, actualPosition)
+      args.callback &&
+        args.callback.value &&
+        args.callback.value(rect, openerRect, containerRect, position, actualPosition)
```
oxcened, line 315: I suggest avoiding logical operations in place of if statements as they make code less clear
```suggestion
      if (args.callback?.value) {
        args.callback.value(rect, openerRect, containerRect, position, actualPosition)
      } 
```

packages/typeahead/src/FzTypeahead.vue
```diff
@@ -49,20 +83,35 @@ const [model, modelModifiers] = defineModel<string, "object">({
...
+  const res = props.selectProps ? props.selectProps?.options : props.options;
+  return new Fuse(res || [], fuseOptions);
 });
 
 const safeInputContainer = computed(() => {
   return opener.value?.containerRef;
 });
 
+const handleInputFocus = (isOpen: boolean, handlePickerClick: () => void) => {
+  !isOpen && handlePickerClick();
```
oxcened, line 111:
```suggestion
  if (!isOpen) handlePickerClick();
```

packages/select/src/components/FzSelectOption.vue
```diff
@@ -34,10 +35,20 @@ const mappedClass = {
...
+  } else {
+    res = [
+      props.selectedValue === props.option.value
+        ? "bg-background-alice-blue text-blue-500"
+        : "bg-white hover:!bg-background-alice-blue text-core-black hover:text-blue-500",
+    ];
+  }
+  res.push(mappedClass[props.size]);
+  return res;
+});
```
oxcened, line 53:
```suggestion
const computedClass = computed(() => {
  const { disabled, readonly, value } = props.option;
  const isSelected = props.selectedValue === value
  return [
    {
      'text-grey-200': disabled,
      'text-core-black' : readonly,
      'bg-background-alice-blue text-blue-500': !disabled && !readonly && isSelected,
      'bg-white hover:!bg-background-alice-blue text-core-black hover:text-blue-500': !disabled && !readonly && !isSelected
    },
    mappedClass[props.size],
  ]
});
```

my proposal

PR #166, opened 2024-10-23
feat(LIB-1340): Handle singlefile in FzUpload

Crish15:
Jira: [LIB-1340](https://fiscozen.atlassian.net/browse/LIB-1340)

### DIFF
--- a/packages/upload/src/FzUpload.vue
+++ b/packages/upload/src/FzUpload.vue
@@ -58,7 +58,7 @@
 </template>
 
 <script setup lang="ts">
-import { onUnmounted, ref } from "vue";
+import { onMounted, onUnmounted, ref } from "vue";
 import { FzUploadProps } from "./types";
 import { FzButton, FzIconButton } from "@fiscozen/button";
 import { FzLink } from "@fiscozen/link";
@@ -68,11 +68,17 @@ const props = withDefaults(defineProps<FzUploadProps>(), {
   buttonLabel: "Carica",
   dragAndDropLabel: "o trascina qui",
 });
-const emit = defineEmits([]);
+const emit = defineEmits(["fzupload:change"]);
 const model = defineModel<File[]>();
 
 const input = ref<HTMLInputElement | null>(null);
 const urlByFileMap = ref(new Map<File, string>());
+
+onMounted(() => {
+  if(model.value && model.value.length > 0 && !props.multiple) {
+    console.warn("[Fiscozen Design System]: FzUpload prop 'multiple' is set to false, but multiple files are provided. Only the first file will be used.");
+  }
+})
 
 onUnmounted(() => {
   urlByFileMap.value.forEach((value) => {
@@ -93,11 +99,16 @@ function handleDrop(event: DragEvent) {
 function handleInputChange(event: Event) {
   if (!input.value?.files) return;
   addFiles([...input.value.files]);
+  emit("fzupload:change", model.value);
   input.value.value = "";
 }
 
 function addFiles(filesToAdd: File[]) {
-  model.value = [...(model.value ?? []), ...filesToAdd];
+  if(props.multiple) {
+    model.value = [...(model.value ?? []), ...filesToAdd];
+  } else {
+    model.value = [...filesToAdd];
+  }
 }
 
 function getFileUrl(file: File) {
--- a/packages/upload/src/FzUpload.vue
+++ b/packages/upload/src/FzUpload.vue
@@ -68,7 +68,9 @@ const props = withDefaults(defineProps<FzUploadProps>(), {
   buttonLabel: "Carica",
   dragAndDropLabel: "o trascina qui",
 });
-const emit = defineEmits(["fzupload:change"]);
+const emit = defineEmits<{ 
+  'fzupload:change': [files: File[]]  
+}>(); 
 
 function modelSetter(value: File[]) {
   if (value.length > 1 && !props.multiple) {
@@ -90,7 +92,9 @@ const urlByFileMap = ref(new Map<File, string>());
 
 onMounted(() => {
   // this call is needed to make sure the model is correctly set when passed as a prop
-  model.value = modelSetter(model.value);
+  nextTick(() => {
+    model.value = modelSetter(model.value);
+  });
 });
 
 onUnmounted(() => {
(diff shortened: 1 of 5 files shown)


packages/upload/src/FzUpload.vue
```diff
@@ -68,12 +68,18 @@ const props = withDefaults(defineProps<FzUploadProps>(), {
...
+const emit = defineEmits(["fzupload:change"]);
 const model = defineModel<File[]>();
 
 const input = ref<HTMLInputElement | null>(null);
 const urlByFileMap = ref(new Map<File, string>());
 
+onMounted(() => {
+  if(model.value && model.value.length > 0 && !props.multiple) {
+    console.warn("[Fiscozen Design System]: FzUpload prop 'multiple' is set to false, but multiple files are provided. Only the first file will be used.");
+  }
```
oxcened, line 80: I don't see any logic that stops me from passing more than one file to the model

packages/upload/src/FzUpload.vue
```diff
@@ -93,11 +99,16 @@ function handleDrop(event: DragEvent) {
 function handleInputChange(event: Event) {
   if (!input.value?.files) return;
   addFiles([...input.value.files]);
+  emit("fzupload:change", model.value);
```
oxcened, line 102: What is this for? I don't think it does what you want, you are emitting the old model

packages/upload/src/FzUpload.vue
```diff
@@ -93,11 +99,16 @@ function handleDrop(event: DragEvent) {
...
   input.value.value = "";
 }
 
 function addFiles(filesToAdd: File[]) {
-  model.value = [...(model.value ?? []), ...filesToAdd];
+  if(props.multiple) {
```
oxcened, line 107: there's some spaces missing here and there, you should run prettier

packages/upload/src/FzUpload.vue
```diff
@@ -93,11 +99,16 @@ function handleDrop(event: DragEvent) {
...
+    model.value = [...(model.value ?? []), ...filesToAdd];
+  } else {
+    model.value = [...filesToAdd];
```
oxcened, line 110: It's not granted that you only have one file. If it comes from the input file then yes it should. But if it's come drag and drop, you have to handle that yourself.

packages/upload/src/FzUpload.vue
Crish15: addFiles assign the new model

packages/upload/src/FzUpload.vue
```diff
@@ -68,12 +68,33 @@ const props = withDefaults(defineProps<FzUploadProps>(), {
...
+const emit = defineEmits(["fzupload:change"]);
+const model = defineModel<File[]>({
+  default: [],
+});
 
 const input = ref<HTMLInputElement | null>(null);
 const urlByFileMap = ref(new Map<File, string>());
 
+onMounted(() => {
+  if (model.value && model.value.length > 0 && !props.multiple) {
```
oxcened, line 80:
```suggestion
  if (model.value?.length > 1 && !props.multiple) {
```

packages/upload/src/FzUpload.vue
```diff
@@ -68,12 +68,33 @@ const props = withDefaults(defineProps<FzUploadProps>(), {
...
+  }
+});
+
+watch(
+  () => model.value,
+  (newFiles: File[]) => {
+    if (newFiles.length > 1 && !props.multiple) {
+      model.value = [newFiles[0]];
+    }
+    emit("fzupload:change", model.value);
```
oxcened, line 94: I think this is redundant, there's already update:modelValue which does the exact same thing

packages/upload/src/FzUpload.vue
```diff
@@ -68,12 +68,33 @@ const props = withDefaults(defineProps<FzUploadProps>(), {
...
+    model.value = [model.value[0]];
+  }
+});
+
+watch(
+  () => model.value,
+  (newFiles: File[]) => {
+    if (newFiles.length > 1 && !props.multiple) {
+      model.value = [newFiles[0]];
+    }
```
oxcened, line 93: I would do this in the setter of the model

packages/upload/src/FzUpload.vue
```diff
@@ -68,12 +68,33 @@ const props = withDefaults(defineProps<FzUploadProps>(), {
...
+});
 
 const input = ref<HTMLInputElement | null>(null);
 const urlByFileMap = ref(new Map<File, string>());
 
+onMounted(() => {
+  if (model.value && model.value.length > 0 && !props.multiple) {
+    console.warn(
+      "[Fiscozen Design System]: FzUpload prop 'multiple' is set to false, but multiple files are provided. Only the first file will be used.",
+    );
```
oxcened, line 83: this warning could also go in the setter, even more so considering you're only checking at mount and never after that

packages/upload/src/FzUpload.vue
Crish15: emitting the change allows the component to be used without passing a model. It is also easier to understand than ‘update:modelValue’.

packages/upload/src/FzUpload.vue
```diff
@@ -68,12 +68,31 @@ const props = withDefaults(defineProps<FzUploadProps>(), {
   buttonLabel: "Carica",
   dragAndDropLabel: "o trascina qui",
 });
-const emit = defineEmits([]);
-const model = defineModel<File[]>();
+const emit = defineEmits(["fzupload:change"]);
```
oxcened, line 71:
```suggestion
const emit = defineEmits<{
  fzupload:change: [files: File[]]
}>();
```

packages/upload/src/FzUpload.vue
```diff
@@ -97,7 +116,13 @@ function handleInputChange(event: Event) {
...
 
 function addFiles(filesToAdd: File[]) {
-  model.value = [...(model.value ?? []), ...filesToAdd];
+  if (props.multiple) {
+    model.value = [...(model.value ?? []), ...filesToAdd];
+  } else {
+    model.value = [filesToAdd[0]];
+  }
+
+  emit("fzupload:change", model.value);
```
oxcened, line 125: this works as is? I thought changing a model was async

Crish15: it should works, but to be safe I changed it

oxcened (approved):
lgtm

PR #170, opened 2024-10-25
feat(LIB-1360): add Label on FzSelect

Crish15:
Jira: [LIB-1360](https://fiscozen.atlassian.net/browse/LIB-1360)

### DIFF
--- a/packages/select/src/FzSelect.vue
+++ b/packages/select/src/FzSelect.vue
@@ -67,23 +67,24 @@
       ref="containerRef"
       test-id="fzselect-options-container"
     >
-      <template
-        v-if="visibleOptions.length"
-        v-for="option in visibleOptions"
-        :key="option.kind === 'label' ? option.label : option.value"
-      >
-        <FzSelectLabel
-          v-if="option.kind === 'label'"
-          :option="option"
-          :size="size"
-        />
-        <FzSelectOption
-          v-else
-          @click="() => handleSelect(option.value)"
-          :option="option"
-          :size="size"
-          :selectedValue="model"
-        />
+      <template v-if="visibleOptions.length">
+        <template
+          v-for="option in visibleOptions"
+          :key="option.kind === 'label' ? option.label : option.value"
+        >
+          <FzSelectLabel
+            v-if="option.kind === 'label'"
+            :option="option"
+            :size="size"
+          />
+          <FzSelectOption
+            v-else
+            @click="() => handleSelect(option.value)"
+            :option="option"
+            :size="size"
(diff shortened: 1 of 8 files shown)


packages/select/src/FzSelect.vue
```diff
@@ -67,10 +67,18 @@
       ref="containerRef"
       test-id="fzselect-options-container"
     >
-      <template v-if="visibleOptions.length">
+      <template
+        v-if="visibleOptions.length"
+        v-for="option in visibleOptions"
```
oxcened, line 72: anti-pattern to mix v-if and v-for

packages/select/src/types.ts
```diff
@@ -77,4 +79,13 @@ export type FzSelectOptionsProps = {
...
   readonly?: boolean;
+  /**
+   * the type of option
+   */
+  kind?: "option";
+};
+
+export type FzSelectLabelProps = {
+  label: string;
+  kind: "label";
```
oxcened, line 90: missing documentation

packages/select/src/components/FzSelectLabel.vue
```diff
@@ -0,0 +1,32 @@
...
+import { computed } from "vue";
+import { FzSelectLabelProps } from "../types";
+
+const props = defineProps<{
+  option: FzSelectLabelProps;
+  size: "sm" | "md" | "lg";
+}>();
+
+const staticClass =
+  "text-grey-400 flex items-center text-ellipsis whitespace-nowrap";
```
oxcened, line 22: is it useful to have this declared on its own?

packages/select/src/components/FzSelectLabel.vue
```diff
@@ -0,0 +1,32 @@
+<template>
+  <button
+    disabled
+    test-id="fzselect-label"
+    :title="option.label"
+    :class="computedClass"
+  >
+    {{ option.label }}
+  </button>
```
oxcened, line 9: why a button element for a label?

packages/select/src/components/FzSelectLabel.vue
Crish15: is cleaner and easier to move into a composable in the future (it share the same style as the action list label)

packages/select/src/components/FzSelectLabel.vue
Crish15: to interrupt the click event. Label or span elements with the disabled attribute do not work. I also tried blocking the click event with some ‘@click.stop’/‘@click.prevent’ modifiers, but they did nothing.

PR #173, opened 2024-11-05
Feat/lib 1274 layering

Trebh:
Jira: [LIB-1274](https://fiscozen.atlassian.net/browse/LIB-1274)

### DIFF
--- a/apps/storybook/package.json
+++ b/apps/storybook/package.json
@@ -60,7 +60,6 @@
     "@chromatic-com/storybook": "^1",
     "@fiscozen/eslint-config": "workspace:^",
     "@fiscozen/prettier-config": "workspace:^",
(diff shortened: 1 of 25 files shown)


Crish15 (commented):
I think it is necessary to update the versions of all packages. Or would you prefer to update the version of the packages with the individual releases? (e.g. with the release of the next new feature on FzSelect the layering of FzSelect is also updated)

Trebh:
> I think it is necessary to update the versions of all packages. Or would you prefer to update the version of the packages with the individual releases? (e.g. with the release of the next new feature on FzSelect the layering of FzSelect is also updated)

I will need to update everything, yes

oxcened (approved):

PR #174, opened 2024-11-06
Feat/lib 1271 fzlayout

Trebh:
Jira: [LIB-1271](https://fiscozen.atlassian.net/browse/LIB-1271)

[LIB-1271]: https://fiscozen.atlassian.net/browse/LIB-1271?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/layout/src/FzLayout.vue
+++ b/packages/layout/src/FzLayout.vue
@@ -54,13 +54,13 @@ const layoutClass = computed(() => {
             res = 'grid-rows-auto'
             break;
         case 'twoColumns':
-            res = 'grid-cols-1 grid-rows-[60px_1fr] md:grid-rows-[1fr_1px] md:grid-cols-2'
+            res = 'grid-cols-1 grid-rows-[60px_1fr] md:grid-rows-1 md:grid-cols-2'
             break;
         case 'leftShoulder':
-            res = 'grid-rows-[40px_160px_60px_1fr] grid-cols-[1fr_1px] lg:grid-rows-[40px_160px_1fr] lg:grid-cols-[260px_1fr]'
+            res = 'grid-rows-[40px_160px_60px_1fr] grid-cols-1 lg:grid-rows-[40px_160px_1fr] lg:grid-cols-[260px_1fr]'
             break;
         case 'rightShoulder':
-            res = 'grid-rows-[40px_160px_60px_1fr] grid-cols-[1fr_1px] lg:grid-rows-[40px_160px_1fr] lg:grid-cols-[1fr_260px]'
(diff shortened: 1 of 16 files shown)


packages/layout/src/__tests__/FzLayout.spec.ts
```diff
@@ -0,0 +1,14 @@
...
+
+describe.concurrent('FzLayout', () => {
+  it('matches snaphost', async ({ expect }) => {
+    const wrapper = mount(FzLayout, {
+      props: {},
+      slots: {}
+    })
+
+    expect(wrapper.html()).toMatchSnapshot()
+  })
```
Crish15, line 13: I would add other tests based on the values the layout can take

packages/layout/src/FzLayout.vue
```diff
@@ -0,0 +1,75 @@
...
+            </div>
+            <div class="col-span-2">
+                <slot name="header"></slot>
+            </div>
+            <div class="col-span-2 lg:col-span-1">
+                <slot></slot>
+            </div>
+            <div class="col-span-2 lg:col-span-1">
+                <slot name="right-shoulder"></slot>
+            </div>
```
oxcened, line 32: I would rather have a single template for each layout, it's more redundant but makes it more clear and easier to maintain
There's also two default slots for `rightShoulderNavbar`

packages/layout/src/FzLayout.vue
```diff
@@ -0,0 +1,75 @@
...
+            res = 'grid-rows-[40px_160px_60px_1fr] grid-cols-1 lg:grid-rows-[40px_160px_1fr] lg:grid-cols-[260px_1fr]'
+            break;
+        case 'rightShoulderNavbar':
+            res = 'grid-rows-[40px_160px_60px_1fr] grid-cols-1 lg:grid-rows-[40px_160px_1fr] lg:grid-cols-[1fr_260px]'
+            break;
+        case 'squares':
+            res = 'grid-cols-1 grid-rows-auto sm:grid-cols-2 sm:grid-rows-auto md:grid-cols-3 md:grid-rows-auto auto-rows-[332px]'
+            break;
+        default:
+            break;
```
oxcened, line 67: I'd just return those strings. no need for res

packages/layout/src/FzLayout.vue
Trebh: I'll fix the double slot. As per the single template, I actually prefer this way at least for "default slot - only" templates.

packages/layout/src/FzLayout.vue
Trebh: I very much prefer functions to have a single exit if possible, except for quick "out" conditions at the very top when needed.

oxcened (approved):

Crish15 (approved):

PR #175, opened 2024-11-06
Feat/lib 1368 icon dropdown

Trebh:
Jira: [LIB-1368](https://fiscozen.atlassian.net/browse/LIB-1368)

### DIFF
--- a/packages/dropdown/src/FzIconDropdown.vue
+++ b/packages/dropdown/src/FzIconDropdown.vue
@@ -1,24 +1,24 @@
 <template>
   <FzDropdown v-bind="props" ref="dropdown">
     <template #opener>
-        <FzIconButton :iconName="iconName" @click="dropdown?.open()" :variant="buttonVariant"/>
+      <FzIconButton :iconName="iconName" @click="dropdown?.open()" :variant="buttonVariant" />
     </template>
   </FzDropdown>
 </template>
 
 <script setup lang="ts">
-import { ComponentPublicInstance, ref } from 'vue';
-import FzDropdown from './FzDropdown.vue';
-import { type FzIconDropdownProps } from './types';
-import { FzIconButton } from '@fiscozen/button';
+import { ref } from 'vue'
+import FzDropdown from './FzDropdown.vue'
+import { type FzIconDropdownProps } from './types'
+import { FzIconButton } from '@fiscozen/button'
 
-const dropdown = ref<ComponentPublicInstance>();
+const dropdown = ref<InstanceType<typeof FzDropdown>>()
 
 const props = withDefaults(defineProps<FzIconDropdownProps>(), {
-    iconName: 'bars',
-    closeOnActionClick: true,
-    align: 'center',
-    teleport: true,
-    buttonVariant: 'secondary'
+  iconName: 'bars',
+  closeOnActionClick: true,
+  align: 'center',
+  teleport: true,
+  buttonVariant: 'secondary'
 })
 </script>
--- a/packages/dropdown/src/FzIconDropdown.vue
+++ b/packages/dropdown/src/FzIconDropdown.vue
@@ -1,17 +1,24 @@
 <template>
-  <FzDropdown v-bind="props">
-    <template #opener="slotProps">
-        <FzIconButton :iconName="iconName" @click="slotProps.isOpen = !slotProps.isOpen"/>
+  <FzDropdown v-bind="props" ref="dropdown">
+    <template #opener>
+        <FzIconButton :iconName="iconName" @click="dropdown?.open()" :variant="buttonVariant"/>
     </template>
   </FzDropdown>
 </template>
 
 <script setup lang="ts">
+import { ComponentPublicInstance, ref } from 'vue';
 import FzDropdown from './FzDropdown.vue';
-import { FzIconDropdownProps } from './types';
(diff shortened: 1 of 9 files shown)


Crish15 (commented):
I would add tests on the new component

Trebh:
> I would add tests on the new component

there is very little to test though, it's a very shallow wrapper on FzDropdown... also, since we will be probably changing test suite I didn't feel like adding more things to migrate afterwards

packages/dropdown/src/types.ts
```diff
@@ -30,19 +32,34 @@ type FzDropdownProps = {
...
+  default(props: { isOpen: boolean }): VNode | VNode[]
   /**
    *
    * Use this to replace the button opener entirely
    */
-  opener(props: { isOpen: boolean }): any
+  opener(props: { isOpen: boolean }): VNode | VNode[]
+}
+
+interface FzIconDropdownProps extends FzDropdownProps {
```
oxcened, line 54: I'm pretty sure that storybook isn't able to document props automatically this way

oxcened:
> > I would add tests on the new component
> 
> there is very little to test though, it's a very shallow wrapper on FzDropdown... also, since we will be probably changing test suite I didn't feel like adding more things to migrate afterwards

I don't know about this, all components should be tested and the fact that "it's very shallow wrapper on FzDropdown" is just an implementation detail that could very well change in the future

Trebh:
> > > I would add tests on the new component
> > 
> > 
> > there is very little to test though, it's a very shallow wrapper on FzDropdown... also, since we will be probably changing test suite I didn't feel like adding more things to migrate afterwards
> 
> I don't know about this, all components should be tested and the fact that "it's a very shallow wrapper on FzDropdown" is just an implementation detail that could very well change in the future

Then when and if it changes, add tests.. no?

packages/dropdown/src/types.ts
Trebh: I tried to make it a `type`, but it just refuses to compile in storybook. I tried in TS playground and works just fine, but here it breaks the compilation, no idea why

oxcened:
> > > > I would add tests on the new component
> > > 
> > > 
> > > there is very little to test though, it's a very shallow wrapper on FzDropdown... also, since we will be probably changing test suite I didn't feel like adding more things to migrate afterwards
> > 
> > 
> > I don't know about this, all components should be tested and the fact that "it's a very shallow wrapper on FzDropdown" is just an implementation detail that could very well change in the future
> 
> Then when and if it changes, add tests.. no?

So based on your reasoning, I should create tests for a component only when changing it. How do I ensure that I didn't break it with my changes if there were no tests to begin with?

The fact that FzDropdown is tested doesn't mean that FzIconDropdown is bug proof. This is a faulty assumption. You're also using FzIcon as well, who tells me that it's working as expected together with FzDropdown?

Trebh:
> > > > > I would add tests on the new component
> > > > 
> > > > 
> > > > there is very little to test though, it's a very shallow wrapper on FzDropdown... also, since we will be probably changing test suite I didn't feel like adding more things to migrate afterwards
> > > 
> > > 
> > > I don't know about this, all components should be tested and the fact that "it's a very shallow wrapper on FzDropdown" is just an implementation detail that could very well change in the future
> > 
> > 
> > Then when and if it changes, add tests.. no?
> 
> So based on your reasoning, I should create tests for a component only when changing it. How do I ensure that I didn't break it with my changes if there were no tests to begin with?
> 
> The fact that FzDropdown is tested doesn't mean that FzIconDropdown is bug proof. This is a faulty assumption. You're also using FzIcon as well, who tells me that it's working as expected together with FzDropdown?

Tests should test what this component does, not what Dropdown or Icon do. Those components will have their own tests.

```
<FzIconButton :iconName="iconName" @click="dropdown?.open()" :variant="buttonVariant"/>
```

This is quite literally the only logic of the component. I am arguing that it is not really worth to test this, that's all.

If you are worried that changing dropdown might break this wrapper, I wil add a snapshot test. Which is in my opinion more damaging that useful since it will "break" FzIconDropdown if you change FzDropdown no matter what, be it a useful change or a disruptive one.

Trebh:
> So based on your reasoning, I should create tests for a component only when changing it. How do I ensure that I didn't break it with my changes if there were no tests to begin with?

I would assume if you are working with a component you are looking at it, and if it has no logic and no "branches" of behavior, no specific things to check other than "it shows" then we should be fine, shouldn't we?

As per the "correct" appearance, our workflow (will) mandate that we trigger a Chromatic build (which will, as we know, detect **all** visual changes) and we will have Q&A on that. Which is the only way to ensure that the component is rendered correctly

packages/dropdown/src/FzIconDropdown.vue
```diff
@@ -0,0 +1,24 @@
+<template>
+  <FzDropdown v-bind="props" ref="dropdown">
+    <template #opener>
+      <FzIconButton :iconName="iconName" @click="dropdown?.open()" :variant="buttonVariant" />
```
oxcened, line 4: what if you provide the open function from the opener slot? It would make it much more handy

Trebh: makes sense. I think I will keep the exposed "open" function anyway as an escape hatch for users that may need that

PR #183, opened 2024-11-20
feat(LIB): left and right click handler on FzInput

Crish15:
For Jira ticket: [LIB-992](https://fiscozen.atlassian.net/browse/LIB-992)

### DIFF
--- a/packages/input/src/types.ts
+++ b/packages/input/src/types.ts
@@ -67,14 +67,16 @@ type FzInputProps = {
    * native maxlength input value
    */
   maxlength?: number;
+
   /**
-   * Handler of left icon click
+   * right icon class
    */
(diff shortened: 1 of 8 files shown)


packages/input/src/types.ts
```diff
@@ -67,6 +67,14 @@ type FzInputProps = {
...
    */
   maxlength?: number;
+  /**
+   * Handler of left icon click
+   */
+  onLeftIconClick?: () => void;
+  /**
+   * Handler of right icon click
+   */
+  onRightIconClick?: () => void;
```
oxcened, line 77: why are these prop and not event emitters?

Trebh (approved):

PR #184, opened 2024-11-21
feat(LIB): Emit events for textarea

Crish15:
Emit blur, focus and paste events

### DIFF
--- a/packages/textarea/src/FzTextarea.vue
+++ b/packages/textarea/src/FzTextarea.vue
@@ -16,6 +16,9 @@
         :minlength
         :maxlength
         :readonly
+        @blur="emit('blur', $event)"
+        @focus="emit('focus', $event)"
+        @paste="emit('paste', $event)"
         v-model="model"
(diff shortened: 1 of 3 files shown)


packages/textarea/src/FzTextarea.vue
```diff
@@ -51,6 +54,8 @@ const props = withDefaults(defineProps<FzTextareaProps>(), {
   resize: "all",
 });
 
+const emit = defineEmits(["focus", "paste", "blur"]);
```
oxcened, line 57: can you use ts so that you can type the params?

Crish15: done in [36b9c02](https://github.com/fiscozen/design_system/pull/184/commits/36b9c0277107da59cd82d18c803bc3e478fdc0bd) 
I did not use the compact syntax because TS threw an error, we should investigate why it does.

oxcened (approved):

PR #186, opened 2024-11-27
Feat/lib 1368 progress component

Trebh:
Jira: [LIB-1369](https://fiscozen.atlassian.net/browse/LIB-1369)

[LIB-1369]: https://fiscozen.atlassian.net/browse/LIB-1369?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storybook/package.json
+++ b/apps/storybook/package.json
@@ -20,7 +20,7 @@
     "chromatic": "npx chromatic --project-token=chpt_1a654273b7f5392"
   },
   "dependencies": {
-    "@awesome.me/kit-8137893ad3": "^1.0.194",
+    "@awesome.me/kit-8137893ad3": "^1.0.197",
     "@fiscozen/actionlist": "workspace:^",
     "@fiscozen/alert": "workspace:^",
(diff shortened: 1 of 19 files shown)


Crish15:
I’m not sure if creating the `Progress` package is the right choice. In reality, `FzProgress` is just a specialized version of `FzIcon`. I think it would be better to create a `FzProgressIcon` component within the `icon` package, without all this boilerplate code that seems unnecessary noise to me. What do you think?

Trebh:
> I’m not sure if creating the `Progress` package is the right choice. In reality, `FzProgress` is just a specialized version of `FzIcon`. I think it would be better to create a `FzProgressIcon` component within the `icon` package, without all this boilerplate code that seems unnecessary noise to me. What do you think?

It was agreed that we would create a progress component to mimic the tree structure made by designers. It will also hold future logic for more complex progress components

PR #187, opened 2024-11-27
Feat/1362 fzcurrency minmax

Trebh:
Jira: [LIB-1362](https://fiscozen.atlassian.net/browse/LIB-1362)

[LIB-1362]: https://fiscozen.atlassian.net/browse/LIB-1362?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/composables/src/composables/useCurrency.ts
+++ b/packages/composables/src/composables/useCurrency.ts
@@ -41,8 +41,16 @@ export const useCurrency = (options: FzUseCurrencyOptions) => {
     }
     let { value } = el
     value = value.replace(/[^0-9,.-]/g, '')
-    setValue(value)
-    const numberValue = vm?.props.nullOnEmpty && value === '' ? null : parse(value)
+    let parsed: number = parse(value);
+    if (options.min && options.min > parsed) {
+      parsed = options.min
+    }
+    if (options.max && options.max < parsed) {
+      parsed = options.max
+    }
+
+    setValue(format(parsed))
+    const numberValue = vm?.props.nullOnEmpty && value === '' ? null : parsed
     emitAmount(Number.isNaN(numberValue) ? 0 : numberValue)
   }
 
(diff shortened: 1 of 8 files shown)


packages/composables/src/composables/useCurrency.ts
```diff
@@ -1,12 +1,12 @@
 import { Ref, watch, getCurrentInstance, computed, ref, nextTick } from 'vue'
 import { FzUseCurrencyOptions } from '../types'
-import { format as formatNumber, parse } from '../utils'
+import { format as formatNumber, parse, roundTo } from '../utils'
 
 export const useCurrency = (options: FzUseCurrencyOptions) => {
   const inputRef: Ref<HTMLInputElement | null | undefined> = ref(null)
   const vm = getCurrentInstance()
```
Crish15, line 7: Accessing the current instance has made this composable highly coupled with the `FzCurrencyInput`: components using this composable must have an `amount` and a `nullOnEmpty` prop, as well as an emitter for the `update:amount` event.  

For now, this is not a major issue, but we should probably create a refactor card to move these composables into the input package.

packages/input/src/FzCurrencyInput.vue
```diff
@@ -107,10 +118,21 @@ onMounted(() => {
...
   });
 });
-const model = defineModel("amount");
+const model = defineModel<number>("amount");
+
+const stepUpDown = (amount: number) => {
+  if (!props.step) {
+    return;
+  }
+  let stepVal = props.forceStep ? roundTo(props.step, model.value || 0) : model.value || 0
```
Crish15, line 127:
```suggestion
  const value = model.value || 0;
  let stepVal = props.forceStep ? roundTo(props.step, value) : value
```

packages/composables/src/composables/useCurrency.ts
Trebh: Correct. The reason I made it this way is I loosely followed what they did in the existing components we used as examples. 
I kind of understnad why it was done this way, it's quite hard to decouple it from components while maintaining reactivity

Trebh: although I would argue this specific PR has little to do with it. If anything, I added utils (which might be a better solution, along with a refactor that moves component logic... well into the component itself I guess)

PR #189, opened 2024-12-03
Feat/lib 1442 fztable

Trebh:
Jira: [LIB-1442](https://fiscozen.atlassian.net/browse/LIB-1442)

Dato che ancora non abbiamo un design definitivo forse per ora mi fermerei qui, intanto abbiamo penso una base su cui costruire, comprensiva anche del nuovo component testing con Cypress.

Se volete ci diamo un occhio insieme

### DIFF
--- a/packages/table/cypress.config.ts
+++ b/packages/table/cypress.config.ts
@@ -1,0 +1,12 @@
+import { defineConfig } from "cypress";
+
+export default defineConfig({
+  viewportWidth: 1000,
+  viewportHeight: 660,
+  component: {
+    devServer: {
+      framework: "vue",
+      bundler: "vite",
+    },
+  },
+});
--- a/packages/table/cypress/support/component.ts
+++ b/packages/table/cypress/support/component.ts
@@ -1,0 +1,66 @@
+// ***********************************************************
+// This example support/component.ts is processed and
+// loaded automatically before your test files.
+//
+// This is a great place to put global configuration and
+// behavior that modifies Cypress.
+//
+// You can change the location of this file or turn off
+// automatically serving support files with the
+// 'supportFile' configuration option.
+//
+// You can read more here:
+// https://on.cypress.io/configuration
+// ***********************************************************
+
+// Import commands.js using ES2015 syntax:
+import './commands'
+import './main.css'
+import { createMemoryHistory, createRouter, RouterLink } from 'vue-router'
+
+// Alternatively you can use CommonJS syntax:
+// require('./commands')
+
+import { mount } from 'cypress/vue'
+import FzTable from '../../src/FzTable.vue'
+
+// Augment the Cypress namespace to include type definitions for
+// your custom command.
+// Alternatively, can be defined in cypress/support/component.d.ts
+// with a <reference path="./component" /> at the top of your spec.
+declare global {
+  namespace Cypress {
+    interface Chainable {
+      mount: typeof mount
+    }
+  }
+}
+
+Cypress.Commands.add('mount', (component, options = {}) => {
+  // Setup options object
+  options.global = options.global || {}
+  options.global.plugins = options.global.plugins || []
+
+  // create router if one is not provided
+  if (!options.router) {
+    options.router = createRouter({
+      routes: [
+        {path: '/foo', component: FzTable}
+      ], history: createMemoryHistory(),
+    })
+  }
+
+  options.global.components = options.global.components || {}
+  options.global.components['RouterLink'] = RouterLink
+  // Add router plugin
+  options.global.plugins.push({
+    install(app) {
+      app.use(options.router)
+    },
+  })
+
+  return mount(component, options)
+})
+
+// Example use:
+// cy.mount(MyComponent)
--- a/packages/table/src/FzTable.cy.ts
+++ b/packages/table/src/FzTable.cy.ts
@@ -1,0 +1,65 @@
+import { h } from 'vue'
+import FzTable from './FzTable.vue'
+import {FzColumn} from '@fiscozen/simple-table'
+
+const sampleObj = {
+  nome: 'Riccardo',
+  cognome: 'Agnoletto',
+  email: 'riccardo.agnoletto@fiscozen.it',
+  phone_number: '123456789'
+}
+
+const items =  [
+  {
+    type: 'button' as const,
+    label: 'Some label'
+  },
+  {
+    type: 'button' as const,
+    label: 'Another label'
+  },
+  {
+    type: 'button' as const,
+    label: 'A slightly longer label'
+  },
+  {
+    type: 'button' as const,
+    label: 'label'
+  },
+];
+
+describe('<FzTable />', () => {
+  it('renders', () => {
+    // see: https://on.cypress.io/mounting-vue
+    const props = {
+      value: (Array(50)).fill({}).map(() => sampleObj),
+      placeholder: 'Nessun valore',
+      actions: {
+        items
+      },
+      pages: 10,
+      activePage: 2
+    }
+    cy.mount(
+      () => {
+        return h(
+          h('div', {class: 'h-[600px] max-w-[800px] pt-16'}, [
+            h(FzTable, props, [
+              h(FzColumn, {header: 'Nome', sticky: 'left'}),
+              h(FzColumn, {header: 'Cognome'}),
+              h(FzColumn, {header: 'Email'}),
+              h(FzColumn, {header: 'Email'}),
+              h(FzColumn, {header: 'Email'}),
+              h(FzColumn, {header: 'Email'}),
+              h(FzColumn, {header: 'Email'}),
+              h(FzColumn, {header: 'Email'}),
+              h(FzColumn, {header: 'Email'}),
+              h(FzColumn, {header: 'Email'}),
+              h(FzColumn, {header: 'Phone Number', field: 'phone_number'})
+            ])
+          ])
+        )
+      }
+    )
+  })
+})
--- a/packages/table/src/FzTable.cy.ts
+++ b/packages/table/src/FzTable.cy.ts
@@ -1,6 +1,7 @@
 import { h } from 'vue'
 import FzTable from './FzTable.vue'
 import {FzColumn} from '@fiscozen/simple-table'
+import { FzTableProps } from './types';
 
 const sampleObj = {
   nome: 'Riccardo',
@@ -28,6 +29,24 @@ const items =  [
   },
 ];
 
+const defaultTemplate = (props: FzTableProps) => () => h(
+  h('div', {class: 'h-[600px] max-w-[800px] pt-16'}, [
+    h(FzTable, props, [
+      h(FzColumn, {header: 'Nome', sticky: 'left'}),
+      h(FzColumn, {header: 'Cognome'}),
+      h(FzColumn, {header: 'Email'}),
+      h(FzColumn, {header: 'Email'}),
+      h(FzColumn, {header: 'Email'}),
+      h(FzColumn, {header: 'Email'}),
+      h(FzColumn, {header: 'Email'}),
+      h(FzColumn, {header: 'Email'}),
+      h(FzColumn, {header: 'Email'}),
+      h(FzColumn, {header: 'Email'}),
+      h(FzColumn, {header: 'Phone Number', field: 'phone_number'})
+    ])
+  ])
+)
+
 describe('<FzTable />', () => {
   it('renders', () => {
     // see: https://on.cypress.io/mounting-vue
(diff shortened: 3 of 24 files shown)


packages/table/src/FzTable.vue
```diff
@@ -0,0 +1,277 @@
...
+
+<template>
+<div class="fz__table overflow-auto h-full w-full">
+  <div
+    :class="[staticClasses, classes]"
+    ref="grid"
+    role="table"
+    :aria-rowcount="value.length"
+    :aria-colcount="columns.length">
+    <template  v-for="(column, index) in columns">
```
oxcened, line 187: c'è un doppio spazio dopo template, farei girare prettier ed eslint sul package

packages/table/src/FzTable.vue
```diff
@@ -0,0 +1,277 @@
...
+      return res
+    })
+})
+
+onMounted(() => {
+  updateClasses()
+
+  if (props.actions?.items.length) {
+
+  }
```
oxcened, line 174: if vuoto?

packages/table/src/FzTable.vue
```diff
@@ -0,0 +1,277 @@
...
+}
+.fz__table .grid.grid-cols-10 {
+  grid-template-columns: repeat(10, minmax(min-content, 1fr))
+}
+.fz__table .grid.grid-cols-11 {
+  grid-template-columns: repeat(11, minmax(min-content, 1fr))
+}
+.fz__table .grid.grid-cols-12 {
+  grid-template-columns: repeat(12, minmax(min-content, 1fr))
+}
```
oxcened, line 276: This is very verbose, you could just use a dynamic style with the repeat number interpolated

packages/table/src/FzTable.vue
```diff
@@ -0,0 +1,277 @@
...
+        v-for="column in columns">
+        {{ row[column.props.field?.toLowerCase() || column.props.header.toLowerCase()] }}
+      </div>
+      <div
+        v-if="actions"
+        :class="[bodyStaticClasses, bodyClasses]">
+        <FzIconDropdown :actions="actions.items"></FzIconDropdown>
+      </div>
+    </template>
+    <div :class="['fz__table__empty', 'h-full mt-80', 'justify-self-center', colspan]" v-else>
```
oxcened, line 211: why an array? a string is good enough

packages/table/src/FzTable.vue
```diff
@@ -0,0 +1,277 @@
...
+
+const headerStaticClasses = ['fz__header__cell','sticky','top-0','z-[2]', 'bg-grey-100', 'px-16', 'h-48', 'font-medium', 'flex', 'justify-start', 'items-center']
+const headerClasses = computed(() => {
+
+})
+
+const bodyStaticClasses = ['fz__body', 'z-[1]', 'px-16', 'h-48', 'bg-core-white', 'flex', 'justify-start', 'items-center', 'min-w-min', 'border-b-1', 'border-grey-100']
+const bodyClasses = computed(() => {
+
+})
```
oxcened, line 134: empty?
I'd just make a single computed for each element, I don't see the need to separate between static and non static

packages/table/src/FzTable.vue
```diff
@@ -0,0 +1,277 @@
...
+    ?.filter((elem) => elem.type === FzColumn)
+    .map((column) => ({
+      props: column.props as FzColumnProps,
+      children: column.children as FzColumnSlots,
+    })) ?? [];
+
+const grid = ref()
+
+const staticClasses = ref(['grid'])
+const classes = computed(() => {
```
oxcened, line 30: same here

packages/table/src/FzTable.vue
```diff
@@ -0,0 +1,277 @@
...
+      <div :class="[headerStaticClasses, headerClasses, getBodyClasses(column, true)]" role="columnheader" aria-sort="none" >
+         {{ column.props.header }} 
+      </div>
+    </template>  
+    <div
+      v-if="actions"
+      :class="['fz__table__header--actions', headerStaticClasses, headerClasses]">
+      Azioni
+    </div>
+    <template v-if="value && value.length" v-for="(row, index) in value">
```
oxcened, line 197: you shouldn't mix v-if and v-for 
you also don't really need this condition as v-for will just not iterate in that case

packages/table/src/FzTable.vue
```diff
@@ -0,0 +1,277 @@
...
+    :aria-rowcount="value.length"
+    :aria-colcount="columns.length">
+    <template  v-for="(column, index) in columns">
+      <div :class="[headerStaticClasses, headerClasses, getBodyClasses(column, true)]" role="columnheader" aria-sort="none" >
+         {{ column.props.header }} 
+      </div>
+    </template>  
+    <div
+      v-if="actions"
+      :class="['fz__table__header--actions', headerStaticClasses, headerClasses]">
```
oxcened, line 194: you don't need an array

packages/table/src/FzTable.vue
```diff
@@ -0,0 +1,277 @@
...
+      Azioni
+    </div>
+    <template v-if="value && value.length" v-for="(row, index) in value">
+      <div
+        :aria-rowindex="index"
+        :class="[bodyStaticClasses, bodyClasses, getBodyClasses(column)]"
```
oxcened, line 200: this is confusing, I would just do a single computed `bodyClasses` and put everything there

packages/table/src/FzTable.vue
```diff
@@ -0,0 +1,277 @@
...
+        role="cell"
+        v-for="column in columns">
+        {{ row[column.props.field?.toLowerCase() || column.props.header.toLowerCase()] }}
```
oxcened, line 203:
```suggestion
        {{ (row[column.props.field] || column.props.header)?.toLowerCase() }
```
but why is the header used as fallback?

packages/table/src/FzTable.vue
```diff
@@ -0,0 +1,277 @@
...
+
+})
+
+const getBodyClasses = (column: {props: FzColumnProps, children: FzColumnSlots}, isHeader?: boolean) => {
+  return {
+    relative: !column.props.sticky,
+    sticky: column.props.sticky,
+    ['left-0 z-[2]']: column.props.sticky === 'left',
+    ['z-[3]']: column.props.sticky && isHeader,
+    ['right-0']: column.props.sticky === 'right'
```
oxcened, line 142:
```suggestion
    'left-0 z-[2]': column.props.sticky === 'left',
    'z-[3]': column.props.sticky && isHeader,
    'right-0': column.props.sticky === 'right'
```

packages/table/src/FzTable.vue
```diff
@@ -0,0 +1,277 @@
...
+const defaultSlot = slots.default?.();
+const columns =
+  defaultSlot
+    ?.filter((elem) => elem.type === FzColumn)
+    .map((column) => ({
+      props: column.props as FzColumnProps,
+      children: column.children as FzColumnSlots,
+    })) ?? [];
+
+const grid = ref()
```
oxcened, line 27: this is not typed

packages/table/src/FzTable.vue
```diff
@@ -0,0 +1,277 @@
...
+}
+
+const updateClasses = () => {
+  const tableWidth = columns.reduce((acc, el, index) => {
+    const colWidth = grid.value.children[index].getBoundingClientRect().width
+    acc+=colWidth
+    return acc
+  }, 0)
+  grid.value.style.width = `${tableWidth}px`
+}
```
oxcened, line 153: why is this needed?

packages/table/cypress/support/component.ts
```diff
@@ -0,0 +1,66 @@
...
+    install(app) {
+      app.use(options.router)
+    },
+  })
+
+  return mount(component, options)
+})
+
+// Example use:
+// cy.mount(MyComponent)
```
oxcened, line 66: I'd remove this

packages/table/src/FzTable.vue
```diff
@@ -0,0 +1,277 @@
...
+        <FzIconDropdown :actions="actions.items"></FzIconDropdown>
+      </div>
+    </template>
+    <div :class="['fz__table__empty', 'h-full mt-80', 'justify-self-center', colspan]" v-else>
+      {{ placeholder ?? "No data available" }}
+    </div>
+  </div>
+  
+</div>
+<div class="fz__table__footer w-full flex flex-row justify-end m-8" v-if="pages && value && value.length">
```
oxcened, line 217:
```suggestion
<div class="fz__table__footer w-full flex flex-row justify-end m-8" v-if="pages && value?.length">
```

packages/table/src/FzTable.vue
Trebh: nope. Sadly, tailwind would not compile those classes

Trebh: This is by design in tailwind, and it is in fact the main drawback of using css grid in my opinion

packages/table/src/FzTable.vue
Trebh: colspan is a computed prop and I didn't want to have both class and :class

packages/table/src/FzTable.vue
```diff
@@ -0,0 +1,277 @@
...
+      res.push('col-span-10') 
+      break;
+    case 11:
+      res.push('col-span-11') 
+      break;
+    case 12:
+      res.push('col-span-12') 
+      break;
+    default:
+      break;
```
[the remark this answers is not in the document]
Trebh: I fear not

packages/table/src/FzTable.vue
Trebh: Wasn't this agreed before to separate static from dynamic? of course I will remove empty ones

packages/table/src/FzTable.vue
Trebh: I think it would be easier to show you, but basically it's needed for the "sticky left or right" part sadly

packages/table/src/FzTable.vue
Trebh: unless you mean we go with custom interpolated `:style` binding, which is possible. I don't like it but it is probably better than this

Crish15: why does tailwind not compile these classes? They are in the documentation https://tailwindcss.com/docs/grid-template-columns

packages/table/src/FzTable.vue
```diff
@@ -0,0 +1,277 @@
...
+
+const grid = ref()
+
+const staticClasses = ref(['grid'])
+const classes = computed(() => {
+  let res: string[] = [];
+
+  const colLength = props.actions ? columns.length + 1 : columns.length
+
+  switch (colLength) {
```
Crish15, line 35:
```suggestion
  res.push(`grid-cols-${colLength}`)
```

packages/table/src/FzTable.vue
```diff
@@ -0,0 +1,277 @@
...
+  }
+  return res;
+})
+
+const colspan = computed(() => {
+  let res: string[] = [];
+
+  const colLength = props.actions ? columns.length + 1 : columns.length
+
+  switch (colLength) {
```
Crish15, line 83:
```suggestion
  res.push(`col-span-${colLength}`)
```

packages/table/src/FzTable.vue
Crish15: `v-for` could also be inside the div without the use of `template` which I find verbose and not useful in this case

packages/table/src/FzTable.cy.ts
```diff
@@ -0,0 +1,81 @@
...
+        items
+      },
+      pages: 10,
+      activePage: 2
+    }
+    cy.mount(defaultTemplate(props))
+    cy.get('.fz__table__empty').should('be.visible')
+    cy.get('.fz__table__empty').should('have.text', props.placeholder)
+  })
+})
```
Crish15, line 81: I would add a test for the sticky position

packages/table/src/FzTable.vue
```diff
@@ -0,0 +1,277 @@
...
+      </div>
+    </template>
+    <div :class="['fz__table__empty', 'h-full mt-80', 'justify-self-center', colspan]" v-else>
+      {{ placeholder ?? "No data available" }}
+    </div>
+  </div>
+  
+</div>
+<div class="fz__table__footer w-full flex flex-row justify-end m-8" v-if="pages && value?.length">
+  <div class="fz__table__pagination flex flex-row justify-between items-center gap-8">
```
Crish15, line 218: I would create a separate componente called `FzTabelPagination`: it will be easier to refactor/move to another package when the designers will create a Pagination component

Crish15 (changes requested):

packages/table/src/FzTable.vue
oxcened: yes, I meant using a custom js in css. It's the only doable way afaik

packages/table/src/FzTable.vue
oxcened: ok but why have all the other classes separated

packages/table/src/FzTable.vue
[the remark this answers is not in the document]
oxcened: if you don't do it with classes yes. This is also not dynamic, what if I need 13 columns

packages/table/src/FzTable.vue
Trebh: @Crish15 I meant that if you interpolate names like grid-col-{variable} tailwind JIT will not work. Of course if we bypass that and go with :style, that's fine. Will do that I think, this honestly just sucks

packages/table/src/FzTable.vue
Trebh: fair enough

packages/table/src/FzTable.vue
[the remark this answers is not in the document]
Trebh: same as above, I have to go with custom style. Though this is actually done in some css frameworks. The reasoning behind this -obviously- not scalable approach is that you should not go above 12 or so columns

packages/table/src/FzTable.vue
Trebh: I'll try, I have the feeling that using bindings this will be called a lot, which could become a problem. I'll try that though

packages/table/cypress.config.ts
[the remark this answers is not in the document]
Trebh: I thought about that, but I am not sure. Having a common config: fine, although we could just keep it in the template. Updating test config might break a lot of stuff so I am not sure having it as a package dep is something we want.

Having a separate "test" package would be a no for me: every package should strive to be independent in the sense that should contain all of its sources and tests. Storybook is a different situation as it is more of a showcase app to me

We will discuss this anyway, don't worry

packages/table/src/FzTable.vue
Trebh: this does not work in tailwind sadly

packages/table/src/FzTable.vue
Crish15: > @Crish15 I meant that if you interpolate names like grid-col-{variable} tailwind JIT will not work. Of course if we bypass that and go with :style, that's fine. Will do that I think, this honestly just sucks

Ok, maybe we can try with this solution? https://stackoverflow.com/questions/67289894/jit-tailwindcss-using-variable-in-bg-not-rendering-color

packages/table/src/FzTable.vue
Crish15: I forgot this small limitation of Tailwind 
image

packages/table/src/FzTable.vue
Trebh: the only solution is to use `:style` binding, if we want to avoid the static classes in this situation

Crish15 (approved):

PR #191, opened 2024-12-05
FzRadioGroup default size to md

Crish15:
FzRadioGroup default size to md


oxcened (approved):

oxcened (commented):
I just noticed that size is still required in the props type though

PR #208, opened 2025-02-24
Feat/lib 1621 fz table ordering

Trebh:
Jira: [LIB-1621](https://fiscozen.atlassian.net/browse/LIB-1621)

[LIB-1621]: https://fiscozen.atlassian.net/browse/LIB-1621?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/table/src/FzTable.vue
+++ b/packages/table/src/FzTable.vue
@@ -192,36 +192,38 @@ const handleOrderingClick = (colProps: FzColumnProps) => {
         >
           Azioni
         </div>
-        <div
-          class="grid grid-cols-subgrid border-b-1 border-solid border-grey-100"
-          v-if="internalValue?.length"
-          v-for="(row, index) in internalValue"
-          :aria-rowindex="index + 1"
-          :style="colSpan"
-          role="row"
-        >
-          <slot :name="`row-${index}`" :columns :data="row" :actions>
-            <FzRow :columns :data="row" :actions> </FzRow>
-          </slot>
-        </div>
-        <div
-          v-else-if="rows && rows.length"
-          :class="[
-            'grid grid-cols-subgrid border-b-1 border-solid border-grey-100',
-            bodyStaticClasses,
-          ]"
-          v-for="(row, index) in rows"
-          :aria-rowindex="index + 1"
-          :style="colSpan"
-          role="row"
-        >
-          <component
-            v-if="row.children?.default"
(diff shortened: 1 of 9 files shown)


packages/table/src/FzTable.vue
```diff
@@ -98,19 +95,67 @@ const totalColumns = computed(() => {
...
 
 <template>
   <div class="m-0 p-0 size-full">
+    <div class="flex flex-row justify-end">
+      <FzInput
+        v-if="props.filterable"
+        data-cy="fztable-search"
+        class="mb-8 max-w-xs"
+        :label="props.searchFilterLabel"
+        @update:modelValue="(e: string) => emit('update:searchTerm', e)"
```
Crish15, line 145:
```suggestion
        @update:modelValue="emit('update:searchTerm', $event)"
```

packages/table/src/FzTable.vue
```diff
@@ -137,43 +194,34 @@ const colSpan = computed(() => ({
         </div>
         <div
           class="grid grid-cols-subgrid border-b-1 border-solid border-grey-100"
-          v-if="value && value.length"
-          v-for="(row, index) in value"
+          v-if="internalValue?.length"
+          v-for="(row, index) in internalValue"
```
Crish15, line 198: It's not recommended to use `v-if` and `v-for` in the same element, we should move the `v-if` in a parent `<template>` tag https://vuejs.org/guide/essentials/conditional.html#v-if-with-v-for

packages/table/src/FzTable.vue
```diff
@@ -137,43 +194,34 @@ const colSpan = computed(() => ({
...
-                    emit('fztable:rowactionclick', actionIndex, row)
-                "
-              ></FzIconDropdown>
-            </div>
+          <slot :name="`row-${index}`" :columns :data="row" :actions>
+            <FzRow :columns :data="row" :actions> </FzRow>
           </slot>
         </div>
+        <div
+          v-else-if="rows && rows.length"
```
Crish15, line 208: It's not recommended to use v-if and v-for in the same element, we should move the v-else-if in a parent <template> tag https://vuejs.org/guide/essentials/conditional.html#v-if-with-v-for

PR #210, opened 2025-03-05
fix: Floating position

Crish15:
Jira: [LIB-1292](https://fiscozen.atlassian.net/browse/LIB-1292)

Sometimes, if an FzSelect with many options opens its dropdown upwards and the dropdown is larger than the screen, the dropdown's position might be calculated incorrectly, resulting in a negative Y value. This causes a space to appear between the FzSelect and the dropdown. This fix corrects this by setting any negative Y or X position to 0, so the dropdown stays on the screen.

After internal discussions, we decided it's best to expose `setPosition` from `FzFloating` and call it in `FzSelect` to recalculate the floating position when the container's maxHeight changes.

[LIB-1292]: https://fiscozen.atlassian.net/browse/LIB-1292?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/composables/package.json
+++ b/packages/composables/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@fiscozen/composables",
-  "version": "0.1.34",
+  "version": "0.1.35",
   "description": "Design System utility composables",
   "scripts": {
     "coverage": "vitest run --coverage",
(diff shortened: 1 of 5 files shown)


Trebh:
Even though this modification might be a "last resort" check to avoid float to go out of screen, I don't think it really addres issue 1292.

There is also another thing to consider: what if the float is simply too big for the container? Are we sure we want to preserve top/left positioning while potentially bottom/right will overflow?

PR #387, opened 2026-05-20
fix(LIB-1901): align FzNavbar CSS-variable defaults with @fiscozen/style pixel tokens

francesca-vago:
The v0.2.0 CSS-variable fallbacks (3rem/14rem/8rem/4rem) targeted stock Tailwind's rem-based spacing scale, but the package targets `@fiscozen/style` where the matching Tailwind classes resolve to pixels (p-12 = 12px, h-56/w-56 = 56px, mr-32/mb-32 = 32px, gap-16 = 16px).

Because the package's `.fz-navbar.{w-56,h-56,…}` rules have higher specificity than the consumer-generated `.w-56` etc. on their own, the rem defaults displaced the pixel values and the navbar rendered ~4× larger than in v0.1.x on consumers without explicit overrides. This restores the historical visual sizing while keeping every CSS variable consumer-overridable.

Jira: [LIB-1901](https://fiscozen.atlassian.net/browse/LIB-1901)

[LIB-1901]: https://fiscozen.atlassian.net/browse/LIB-1901?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/navbar/src/FzNavbar.vue
+++ b/packages/navbar/src/FzNavbar.vue
@@ -105,15 +105,21 @@ function handleMenuButtonClick() {
 
 <style>
 /*
- * CSS custom properties — defaults match the previous Tailwind utility values
- * (p-12 = 3rem, shadow, z-10, h-56/w-56 = 14rem, mr-32/mb-32 = 8rem, gap-16 = 4rem).
- * Consumers can override per-instance via inline style or scoped CSS to slim
- * the navbar without having to rely on `!important` resets.
+ * CSS custom properties — defaults are pinned to the `@fiscozen/style` pixel
+ * spacing tokens that match the Tailwind utility classes used in the template
+ * (p-12 = 12px, h-56/w-56 = 56px, mr-32/mb-32 = 32px, gap-16 = 16px). The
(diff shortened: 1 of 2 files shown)


packages/navbar/src/FzNavbar.vue
```diff
@@ -155,15 +161,15 @@ function handleMenuButtonClick() {
...
   );
   padding-left: calc(
-    var(--fz-navbar-padding, 3rem) +
+    var(--fz-navbar-padding, 12px) +
       max(env(safe-area-inset-left, 0px), var(--safe-area-inset-left, 0px))
   );
   padding-right: calc(
-    var(--fz-navbar-padding, 3rem) +
+    var(--fz-navbar-padding, 12px) +
       max(env(safe-area-inset-right, 0px), var(--safe-area-inset-right, 0px))
```
sergioterrasifiscozen, line 173: Queste tengono conto anche dell'app mobile, giusto?

PR #417, opened 2026-07-14
feat(layout): FzDetailTemplate — backoffice detail-page layout (LIB-2695)

Trebh:
Jira: [LIB-2695](https://fiscozen.atlassian.net/browse/LIB-2695)

## What

Adds `FzDetailTemplate` — the presentation-only backoffice **detail-page** layout, and the sibling of `FzListTemplate` (#416): where a list page pairs a *filter* rail with a table, a detail page pairs a persistent **summary/context sidebar** (the record's identity, status, meta and actions) with the record's content (typically `FzTabs`/`FzCard`s). It extracts the shape the backoffice `*Detail.vue` pages hand-roll today — `FzLayout leftShoulder` with a `#sidebar` slot (CU, Intrastat, LIPE, Welfare) and a raw `flex` row (VAT declaration, Model770) — so they converge on one responsive layout (reference: *dettaglio dichiarazione IVA*).

It composes the region molecules (`FzLayoutAside` + `FzLayoutMain`) with **document scroll** rather than wrapping `FzLayout leftShoulder` (which forces `100vh` mobile tracks and independent per-region scroll — exactly what the hand-rolled pages avoid). Being a *page-content* template, it does **not** force a viewport height or apply root safe-area insets; the shell owns those.

## API

| Prop | Type | Default | Notes |
| --- | --- | --- | --- |
| `sidebarLabel` | `string` | — | accessible name (`aria-label`) for the sidebar's `complementary` landmark |
| `mainAs` | `'main' \| 'div'` | `'main'` | set `'div'` when nested in a shell that already owns `<main>` |

Slots: `banner` (full-width, above), `sidebar` (the summary rail, a `complementary` landmark, `md:w-[340px]`), `header` (a title/actions toolbar, deliberately **not** a banner landmark), default (the detail body).

No `sidebarPosition` prop: every real BO detail sidebar is a left rail, so a `left`/`right` variant would be speculative (RFC §10 guardrail) — trivially additive later. Additive — no change to the `FzLayout` grid API or the other templates. Changeset: `@fiscozen/layout` **minor**.

## Base / stacking

Branched off `feat/LIB-2693-page-templates-extraction` (#414) as a **sibling of #415 and #416, not stacked on either** — the detail template consumes only the region molecules from #414, not `FzAppTemplate`/#415 (a page-content layout renders *inside* the shell, it doesn't compose it) nor `FzListTemplate`/#416 (the two are independent page-content siblings). Base auto-retargets to `main` when #414 merges.

## Tests

- **Unit — 125 passed** (6 files; 14 new): rendering/structure, no forced viewport height, optional regions (banner / sidebar-as-complementary / header-not-banner), `sidebarLabel`, `mainAs` main/div, attribute forwarding onto the root, no events.
- **Storybook play (Playwright)**: `Default` / `WithoutSidebar` / `NestedInShell` — the last asserts the `mainAs="div"` landmark de-duplication (no nested `<main>`). Example content is DS components only (`FzAlert` / `FzCard` / `FzBadge` / `FzDivider` / `FzButton` / `FzTabs`).
- Full build (`vue-tsc` + vite) green; visual check at 1280px + 390px confirms the rail-beside-body geometry and mobile summary-first stacking.

## Follow-up (out of scope, separate BO card under RT-2054)

Migrating the BO `*Detail.vue` pages onto `FzDetailTemplate` (converging the two hand-rolled sidebar/body variants) and reconciling who owns `<main>` between `FzAppTemplate` and this template live app-side in `fiscozen-app`.

[LIB-2695]: https://fiscozen.atlassian.net/browse/LIB-2695?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/layout/src/types.ts
```diff
@@ -187,6 +187,56 @@ type FzFocusTemplateSlots = {
...
+  sidebarLabel?: string
+  /**
+   * The element the main content region renders as. Defaults to `main` so the
+   * detail body — the page's primary content — is a `main` landmark. Set to
+   * `div` when the template is composed inside a shell that already renders a
+   * `<main>` (e.g. `FzAppTemplate`'s default slot), to avoid nested `main`
+   * landmarks.
+   * @default 'main'
+   */
+  mainAs?: 'main' | 'div'
```
sergioterrasifiscozen, line 220: Possiamo valutare l'uso di `tag?: keyof HTMLElementTagNameMap`?

sergioterrasifiscozen (approved):

packages/layout/src/types.ts
sergioterrasifiscozen: Mi rispondo io: no.

PR #426, opened 2026-08-03
fix(datepicker): budget the calendar height and honour autoPosition

panicofr:
Jira: [LIB-2814](https://fiscozen.atlassian.net/browse/LIB-2814)

Addresses [HD-25540](https://fiscozen.atlassian.net/browse/HD-25540): from the app, the payment-date calendar is cut off at the bottom of the screen and cannot be scrolled to.

## Root cause (reproduced in the app's own WebView)

The trigger is the **Android system font scale**. Android WebViews honour it; Chrome on Android uses its own text-scaling preference, default 100% — which is exactly why the customer only sees this in the app and why it never reproduced in a browser.

Measured in the real Capacitor WebView (`FiscozenApp/1.0.3`, Chrome 134), 360×728, `font_scale 1.15`, inside `FzConfirmDialog`:

```
field 450..473 · spaceBelow 255 · spaceAbove 450
menu h284 494..778 · flippedAbove: NO · CUTOFF: 50px · menuScrollable: no
```

At `font_scale 1.0` the same probe fits comfortably. Ruled out with measurements in that WebView: the keyboard (the field doesn't raise the IME), safe-area/edge-to-edge (`innerHeight == visualViewport.height`, insets all `0px`), and display size (customer is on the default).

## The two defects fixed

**1. No height budget.** The menu was laid out at its intrinsic height wherever VueDatePicker put it — it does not scroll, and the dialog body behind it does not scroll it into view. So anything past the fold was simply unreachable.

It now gets a `max-height` budgeted from the space actually available on the roomier side of the field, measured against `visualViewport` (so an open keyboard counts too), plus `overflow-y: auto`. Recomputed on resize/scroll/visual-viewport change while open; cleared on close and on unmount. CSS fallback `calc(100dvh - 16px)` if the measurement never runs.

This is the mechanism-independent guarantee: Android goes up to `font_scale 2.0`, and some skins stack a display-size multiplier on top, so there are configurations where **no** placement fits beside the field. Only an internal scroll keeps the calendar usable.

**2. `autoPosition` was dead API.** Documented, defaulting to `true`, and `delete`d on the way to VueDatePicker — which replaced it in v12 with the Floating UI middlewares `floating.flip` / `floating.shift`. `autoPosition: false` did nothing at all. Now translated, with an explicit `floating.flip` / `floating.shift` still authoritative.

## Notes for review

- The budget is written as a custom property on `document.documentElement`, not inline on the menu: the menu is teleported to `body`, and only one calendar can be open at a time (opening another closes the first). That assumption is stated in the code comment.
- I added `@open` on the inner `VueDatePicker`. `open` is a documented pass-through emit; Vue merges the explicit handler with the parent's fallthrough listener, so consumers still receive it.
- **This does not change where the calendar is placed.** Why it doesn't flip above the field despite 450px of room is still open — my hypothesis is that Floating UI measures the menu before the calendar has grown and never recomputes. The height budget makes the outcome non-blocking either way, and I'd rather ship that than a placement change I can't yet justify.
- Deliberately **not** done: neutralising the font scale via `WebSettings.setTextZoom(100)` or `text-size-adjust: none`. It would make the report disappear while overriding an accessibility preference.

## Tests

11 new cases (4 on the `autoPosition` translation, 7 on the budget: below/above/viewport-clamped/never-negative/cleared-on-close/detached-on-unmount). `@fiscozen/datepicker` green at **121/121**.

One test-hygiene note: the "late resize must not resurrect the budget" assertion I first wrote passed in isolation but failed in the full file — other tests in this spec open a calendar and stay mounted, so they recompute the shared custom property. It now asserts listener detachment via `removeEventListener` instead, which is deterministic.

## Still open on LIB-2814

Follow-ups recorded on the card, not in this PR: the mobile bottom-sheet presentation below `sm` (design call), moving calendar cell sizing from `px` to `rem`/`em`, and adding font scale to the QA matrix.

[LIB-2814]: https://fiscozen.atlassian.net/browse/LIB-2814?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HD-25540]: https://fiscozen.atlassian.net/browse/HD-25540?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/datepicker-menu-height-budget.md
+++ b/.changeset/datepicker-menu-height-budget.md
@@ -1,0 +1,15 @@
+---
+"@fiscozen/datepicker": patch
+---
+
+FzDatepicker: keep the calendar reachable when the system font scale grows the page, and make `autoPosition` do something
+
+Two independent defects, both surfaced by a customer report of the calendar being cut off at the bottom of the screen with no way to scroll to it.
+
+**The calendar had no height budget.** It was laid out at its intrinsic height wherever VueDatePicker placed it, so it could hang below the visible area unreachable: the menu does not scroll, and the surface behind it — usually a dialog body — does not scroll it into view either. Android WebViews honour the *system* font scale, so a user with larger text gets a taller form, a date field pushed further down, and no longer enough room beneath it for the calendar. Reproduced in the app's WebView at a 360x728 viewport with font scale 1.15: the field sits at 450..473, the calendar opens below it anyway and 50px of it falls outside the screen. At the largest accessibility font scales no placement fits at all, so an internal scroll is the only thing that keeps the calendar usable.
+
+The menu now gets a `max-height` budgeted from the space actually available on the roomier side of the field — measured against `visualViewport`, so an open keyboard counts too — plus `overflow-y: auto`. The budget is recomputed on resize, scroll and visual-viewport changes while the calendar is open, and cleared when it closes. A CSS fallback of `calc(100dvh - 16px)` keeps the calendar inside the viewport even if the measurement never runs.
+
+**`autoPosition` was dead API.** The prop is documented and defaults to `true`, but it was deleted on the way to VueDatePicker, which replaced it in v12 with the Floating UI middlewares `floating.flip` / `floating.shift`. Setting `autoPosition: false` therefore had no effect whatsoever. It is now translated into `flip: false` / `shift: false`, and an explicit `floating.flip` / `floating.shift` still wins so the finer-grained option stays authoritative.
+
+Note: this does not change *where* the calendar is placed. A separate question — why the menu does not flip above the field even when there is room — is still open; the height budget makes the outcome non-blocking either way.
--- a/packages/datepicker/src/FzDatepicker.vue
+++ b/packages/datepicker/src/FzDatepicker.vue
@@ -108,7 +108,9 @@ const applyMenuHeightBudget = () => {
   const visibleTop = visible?.offsetTop ?? 0
   const visibleHeight = visible?.height ?? window.innerHeight
 
-  // Without a field to measure against, fall back to the whole visible area.
+  // The visible area is the widest defensible budget: it bounds the menu even when the
+  // field cannot be measured (not yet rendered, or already gone), which is better than
+  // leaving the menu unbounded.
   let budget = visibleHeight - MENU_VIEWPORT_MARGIN * 2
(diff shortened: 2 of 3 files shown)


francesca-vago (approved):

Trebh (approved):

panicofr:
Follow-up pushed in `8b510b26` addressing both Aikido comments.

**Layout thrashing (`FzDatepicker.vue:104`) — fixed.** The comment was right and this was the more serious of the two. The budget was recomputed straight from `resize`, `visualViewport.resize` and a **capturing** `scroll` listener, so it also fired for scrolling in any container between the field and the document — each call doing a forced layout read (`getBoundingClientRect`) immediately followed by a style write on `documentElement`. That interleaves reads and writes on a hot path and thrashes layout on exactly the low-end mobile devices this cap exists for.

Event-driven recomputes now go through a `requestAnimationFrame` gate, so a burst collapses into one measurement per frame, and any pending frame is cancelled on teardown. This mirrors how `FzFloating` already schedules its own repositioning, so the two components behave consistently. The budget is still applied **synchronously on open** — the menu is about to be shown and must not render unbounded and settle a frame later.

Aikido's canned "How do I fix it?" text (StringBuilder, regex caching, N+1 queries) was irrelevant here, but the finding itself was sound.

**Redundant comment (`FzDatepicker.vue:111`) — fixed.** Replaced with the rationale for *why* the visible area is the right budget when the field can't be measured, rather than restating the `if (input)` below it.

New test asserts 20 scroll events produce exactly one layout read. Worth flagging how it's written: it captures and flushes frames by hand instead of awaiting a real one, because the snapshot block earlier in this spec installs fake timers and leaves its pickers mounted, so real rAF timing isn't dependable there. Counting reads on this picker's own field keeps the assertion immune to those other instances. That's the second time cross-test coupling in this spec bit an assertion I wrote — it may be worth a separate cleanup card for the mounted-and-never-unmounted pickers.

Suite green at 122/122.

@Trebh @francesca-vago — you've both already approved, so flagging that this adds a commit worth a glance.

PR #430, opened 2026-08-04
fix(LIB-2828): budget the calendar from the menu edge, not the field edge

panicofr:
Jira: [LIB-2828](https://fiscozen.atlassian.net/browse/LIB-2828)

The calendar's height budget was measured from the bottom (or top) of the **field**, but VueDatePicker separates field and menu by its own `offset`, plus arrow and padding. That gap was handed back as if it were usable space, so whenever the budget bound, the calendar overflowed by exactly `gap - margin`.

Measured on a 360x640 viewport with the field at 505..525 and the menu opening at 547 — a 22px gap:

| | budget | real space | cut off |
|---|---|---|---|
| from the field edge | `640 - 525 - 8` = **107px** | 85px | **14px** |
| from the menu edge | `640 - 547 - 8` = **85px** | 85px | **0** |

The 14px is deterministic: the 22px gap minus the 8px margin. Confirmed against the shipped bundle with a DevTools override before writing the fix — menu `547..632`, `cutOffBelow: 0`.

## Why measuring from the menu is stable

The edge we measure from is the one Floating UI holds still: it anchors the menu's top when placed below the field and its bottom when placed above. Shrinking the height therefore never moves that edge, so the budget can't oscillate. The menu's rect was already being read to decide which side it's on, so this is a few lines.

The fallback path is deliberately unchanged: before VueDatePicker has rendered the menu there is no menu box to measure, so the field stays the reference and the budget stays generous until the post-render recompute refines it.

## Scope, after 3.2.12

With the arrow fix ([LIB-2829](https://fiscozen.atlassian.net/browse/LIB-2829), #429) the calendar now flips above the field when there's no room below, so this budget is **no longer** what keeps [HD-25540](https://fiscozen.atlassian.net/browse/HD-25540) usable. It's the fallback for the case where *neither* side has room — which is what happens at the largest accessibility font scales, where no popover fits beside the field at all. It matters there, and the arithmetic being right matters whenever it binds.

## Tests

131 pass. The two LIB-2827 expectations that encoded the field-edge formula are updated, and both sides now assert against an explicit non-zero gap (below: `620 - 546 - 8`; above: menu bottom at 482, so `482 - 8`). LIB-2827's regression guard — budget the side the menu is *on*, not the roomier one — is kept as its own test.

## Known gap, not addressed here

The budget refinement and the `ResizeObserver` are both attached inside a `requestAnimationFrame` callback, which doesn't run in a hidden tab. A calendar opened while the tab is in the background keeps the fallback budget and never self-corrects, because the observer is never attached either. Worth a follow-up: attach the observer without depending on rAF and recompute on `visibilitychange`.

[LIB-2828]: https://fiscozen.atlassian.net/browse/LIB-2828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LIB-2829]: https://fiscozen.atlassian.net/browse/LIB-2829?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/datepicker/src/FzDatepicker.vue
+++ b/packages/datepicker/src/FzDatepicker.vue
@@ -114,8 +114,6 @@ const applyMenuHeightBudget = () => {
   let budget = visibleHeight - MENU_VIEWPORT_MARGIN * 2
   if (input) {
     const field = input.getBoundingClientRect()
-    const below = visibleTop + visibleHeight - field.bottom - MENU_VIEWPORT_MARGIN
-    const above = field.top - visibleTop - MENU_VIEWPORT_MARGIN
 
     // Budget the side the menu is *actually* on. An earlier version budgeted the roomier
     // side, assuming VueDatePicker would place the calendar there — it does not always, and
@@ -129,12 +127,32 @@ const applyMenuHeightBudget = () => {
       (el) => el.getBoundingClientRect().height > 0
     )
     if (menu) {
-      const placedBelow = menu.getBoundingClientRect().top >= field.bottom - 1
-      budget = Math.min(budget, placedBelow ? below : above)
+      const menuRect = menu.getBoundingClientRect()
+      const placedBelow = menuRect.top >= field.bottom - 1
+
+      // Measure from the menu's own edge rather than the field's. VueDatePicker separates
+      // the two by its `offset` (plus arrow and padding), and measuring from the field hands
+      // that gap back as if it were usable space, so the calendar overflowed by exactly
+      // `gap - margin` every time. Measured on a 360x640 viewport with the field at 505..525
+      // and the menu opening at 547: a 107px budget (640 - 525 - 8) against 85px of real
+      // space, cut off by 14px — deterministically the 22px gap minus the 8px margin.
+      //
+      // The edge we measure from is the one Floating UI holds still: it anchors the menu's
+      // top when placed below and its bottom when placed above. Shrinking the height
+      // therefore never moves that edge, so the budget cannot oscillate.
+      budget = Math.min(
+        budget,
+        placedBelow
+          ? visibleTop + visibleHeight - menuRect.top - MENU_VIEWPORT_MARGIN
+          : menuRect.bottom - visibleTop - MENU_VIEWPORT_MARGIN
+      )
     } else {
-      // First run happens before VueDatePicker has rendered the menu, so the side is not
-      // known yet. Stay generous rather than flash a needlessly short calendar; the
-      // post-render recompute below refines it.
+      // First run happens before VueDatePicker has rendered the menu, so neither the side
+      // nor the gap is known yet. The field is the only reference available: stay generous
+      // rather than flash a needlessly short calendar, and let the post-render recompute
+      // below refine it against the real menu box.
+      const below = visibleTop + visibleHeight - field.bottom - MENU_VIEWPORT_MARGIN
+      const above = field.top - visibleTop - MENU_VIEWPORT_MARGIN
       budget = Math.min(budget, Math.max(below, above))
     }
   }
--- a/packages/datepicker/src/__tests__/FzDatepicker.spec.ts
+++ b/packages/datepicker/src/__tests__/FzDatepicker.spec.ts
@@ -1927,27 +1927,43 @@ describe('calendar height budget (LIB-2814)', () => {
   }
   const removeMenu = () => document.querySelector('.dp__menu')?.remove()
 
-  it('budgets the side the menu is actually on, not the roomier one', () => {
+  it('measures from the menu edge, not the field, when the calendar opens below', () => {
     const wrapper = mountPicker()
-    // HD-25540 geometry measured on the ephemeral env: field low, 96px below it, 504 above,
-    // and VueDatePicker places the calendar BELOW anyway. Budgeting the roomier side left
-    // the cap at 496px against 88px of real space, so it never bound and the calendar was
-    // cut off by 250px with no row reachable (LIB-2827).
+    // Production geometry (LIB-2828): field 504..524 with the menu opening at 546, so
+    // VueDatePicker's offset leaves a 22px gap between them. Measuring from the field hands
+    // that gap back as usable space, and the calendar overflows by exactly gap - margin.
     setViewportHeight(620)
     stubField(wrapper, 504, 524)
-    stubMenu(546, 324) // below the field
+    stubMenu(546, 324)
     openMenu(wrapper)
-    expect(document.documentElement.style.getPropertyValue(VAR)).toBe(`${620 - 524 - MARGIN}px`)
+    expect(document.documentElement.style.getPropertyValue(VAR)).toBe(`${620 - 546 - MARGIN}px`)
     removeMenu()
   })
 
-  it('budgets the space above when the menu is placed above the field', () => {
+  it('measures from the menu edge when the calendar opens above', () => {
     const wrapper = mountPicker()
+    // Mirror case: above the field the edge Floating UI holds still is the menu's *bottom*,
+    // and the same 22px gap sits between it and the field.
     setViewportHeight(620)
     stubField(wrapper, 504, 524)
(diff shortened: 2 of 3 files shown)


packages/datepicker/src/FzDatepicker.vue
```diff
@@ -129,12 +127,32 @@ const applyMenuHeightBudget = () => {
       (el) => el.getBoundingClientRect().height > 0
     )
     if (menu) {
-      const placedBelow = menu.getBoundingClientRect().top >= field.bottom - 1
-      budget = Math.min(budget, placedBelow ? below : above)
+      const menuRect = menu.getBoundingClientRect()
+      const placedBelow = menuRect.top >= field.bottom - 1
```
francesca-vago, line 131: **The side is decided from a rect that is still mid-transition — and with the new formula a wrong decision silently disables the cap.**

`.dp__menu` is the root of the transitioned component and VueDatePicker's `transitions` default is `true`, so on enter it carries `dp-menu-appear-bottom-enter-from` → `transform: translateY(calc(var(--dp-transition-length) * -1))`, i.e. **−22px**. Vue removes that class on its own `nextFrame()` (a double `requestAnimationFrame`), while the rAF in `handleMenuOpen` is queued earlier — synchronously, on `open` — so it runs first, in the same frame, with the transform still applied. A menu placed below therefore measures at `field.bottom + gap − 22`.

That is survivable when the budget **binds**: `max-height` changes the menu's height → the `ResizeObserver` fires → `scheduleMenuHeightBudget` re-measures a frame later → the height changes again → and the loop keeps re-measuring until the 100ms transform has decayed. That is why your DevTools measurement settles on the correct 85px rather than 107px.

It is *not* survivable when the budget does not bind, because then nothing changes size and nothing ever re-measures:

```
viewport 400 · field 200..220 · gap 10px (field with no help text) · calendar 324px
real menu 230..554   →   measured mid-transition: 208..532
placedBelow: 208 >= 219 ?  NO  →  "above"
budget = menuRect.bottom - 8 = 524   →   never binds
→ calendar uncapped, 154px off-screen and unreachable, for the whole session
```

That is LIB-2827's failure mode again, in exactly the scenario this PR says it still cares about (neither side has room). The discriminant is the size of the gap: Floating UI's `offset` (10px) plus the distance from the `<input>`'s bottom edge to the bottom of the element Floating UI actually anchors to — which is `DpInput`'s root, i.e. the whole `FzInput` including the help/error row. Your production case had 22px, so `524 >= 523` passes by 1px. A field with no help text and no error has a gap of ~10px and picks the wrong side.

Comparing midpoints instead of edges is immune to a ±22px displacement at any realistic menu height:

```suggestion
      const placedBelow = menuRect.top + menuRect.height / 2 >= field.top + field.height / 2
```

In the example above that is 370 vs 210 → below, with margin to spare instead of 1px.

Caveat on my own claim: I could not confirm the frame ordering in jsdom, because the menu never renders there. The mechanism is derived from Vue's transition implementation and from the fact that your 107-vs-85 measurement is exactly the 22px gap. It wants the browser-level check described in the review summary.

packages/datepicker/src/FzDatepicker.vue
```diff
@@ -129,12 +127,32 @@ const applyMenuHeightBudget = () => {
...
+      // Measure from the menu's own edge rather than the field's. VueDatePicker separates
+      // the two by its `offset` (plus arrow and padding), and measuring from the field hands
+      // that gap back as if it were usable space, so the calendar overflowed by exactly
+      // `gap - margin` every time. Measured on a 360x640 viewport with the field at 505..525
+      // and the menu opening at 547: a 107px budget (640 - 525 - 8) against 85px of real
+      // space, cut off by 14px — deterministically the 22px gap minus the 8px margin.
+      //
+      // The edge we measure from is the one Floating UI holds still: it anchors the menu's
+      // top when placed below and its bottom when placed above. Shrinking the height
+      // therefore never moves that edge, so the budget cannot oscillate.
```
francesca-vago, line 142: Small thing, but it is load-bearing for the anti-oscillation argument, so worth stating precisely: Floating UI does not *hold* the bottom edge still when the menu is placed above — it writes `top`, so capping the height moves the bottom **up**. What restores it is Floating UI's own recompute (`autoUpdate`'s `ResizeObserver` on the floating element) re-anchoring the bottom to `reference.top − offset`.

This code is safe only because `scheduleMenuHeightBudget` defers to the next frame, so by the time you re-measure, Floating UI's reposition (a microtask chain) has already landed. Measured synchronously inside an observer callback, the same formula *would* oscillate. Worth a clause in the comment, since the next reader will take "cannot oscillate" as unconditional.

packages/datepicker/src/__tests__/FzDatepicker.spec.ts
```diff
@@ -1927,27 +1927,43 @@ describe('calendar height budget (LIB-2814)', () => {
...
+    // places the calendar BELOW anyway. Budgeting the roomier side left the cap at 496px
+    // against real space in the eighties, so it never bound and the calendar was cut off
+    // with no row reachable.
+    setViewportHeight(620)
+    stubField(wrapper, 504, 524)
+    stubMenu(546, 324)
+    openMenu(wrapper)
+    const budget = parseInt(document.documentElement.style.getPropertyValue(VAR), 10)
+    expect(budget).toBeLessThan(504 - MARGIN)
     removeMenu()
```
francesca-vago, line 1967: `removeMenu()` at the end of each test body rather than in `afterEach` means a failing assertion leaks the stub `.dp__menu` into the rest of the file — and it is three sites now. That is the same fragility #428's description called out ("a leftover menu from another spec in this file was steering the side detection"). One line in `afterEach` and `stubMenu`/`removeMenu` stop being order-dependent.

Related: `openMenu()` leaves the post-open `requestAnimationFrame` pending and nothing cancels it (`stopTrackingMenuHeight` only tracks `scheduledFrame`), so it fires during some later test's `await` and writes the global custom property then. The file already has the manual frame-capture pattern in the coalescing test — reusing it here would remove that leak *and* let these tests assert the post-render refinement, which right now is only asserted in a comment.

francesca-vago (approved):
Verified the arithmetic and ran the suite locally: **131 tests, 130 pass** — the only failure is the pre-existing `data-v` snapshot mismatch, which fails identically on `main`. Both new expectations match the formula (`620 − 546 − 8` for below; menu bottom at 482 → `482 − 8` for above), and moving `below`/`above` into the fallback branch leaves no dead code behind. I like that this one was measured against the shipped bundle *before* the fix was written.

Approving. Two notes inline: one on the side detection (not introduced here, but this PR gives it a new failure mode), one on the anti-oscillation argument.

### The coverage gap is now the expensive part

This budget has been rewritten three times in a day (LIB-2814 → LIB-2827 → LIB-2828), and each round fixed a real arithmetic error that no unit test could have caught. In jsdom every rect is zero, so `placement`/`y` never change, `shouldRender` stays `false`, and the menu is never rendered at all — which is why these tests have to fabricate a `div.dp__menu` and stub its rect. They pin the formula; they cannot see the geometry.

`PlacementBottomStart` / `PlacementBottomEnd` already exist in `Datepicker.stories.ts`. A variant with a short viewport, a field **without help text**, and a single assertion:

```ts
await expect(calendar.getBoundingClientRect().bottom).toBeLessThanOrEqual(window.innerHeight)
```

would have caught all three defects — the wrong side, the arrow suppressing flip, and the gap counted twice — and would catch the one described inline. That is worth more than the next unit test.

Unrelated heads-up while you are in there: `PlacementBottomStart` asserts `calendarRect.top >= inputRect.bottom - 1`. Now that flip actually works (#429), that assertion holds only as long as the story leaves room below the field — it will fail legitimately in a short viewport.

PR #431, opened 2026-08-05
fix(LIB-2831): make a flipped panel's gap follow the side it landed on

panicofr:
Jira: [LIB-2831](https://fiscozen.atlassian.net/browse/LIB-2831)

The opener-aware vertical correction from [LIB-2813](https://fiscozen.atlassian.net/browse/LIB-2813) (composables 1.1.2) moves a panel to the other side of its opener when it doesn't fit — but the spacing didn't move with it. `FzFloating` derives its gap class from the resolved position, so a `top` panel flipped below its opener kept `mb-4`: margin on the edge facing *away* from the opener.

## Why this isn't cosmetic

The gap class is load-bearing. The position calculators take the gap from the element's own margin:

- `bottom*` → `y = opener.bottom`, and `margin-top` shifts the fixed element down to create the gap
- `top*` → `y = opener.top - margins.bottom`, gap subtracted outright

The correction, meanwhile, spaced a flip by `VIEWPORT_MARGIN` (8). So a flipped panel sat **8px** from its opener where a natively-placed one on that side sits **16px**, while carrying a `margin-bottom` that does nothing to a fixed element positioned by `top`.

## Change

Both halves now come from the same place:

- the flip is spaced by the gap the requested side declares in its margin class — above the opener the gap is in the arithmetic, below it `y` sits on the opener's edge and the flipped side's `margin-top` supplies it, exactly matching the `bottom*` calculator;
- `applyBoundaryCorrections` reports the side it landed on (`flippedTo`), and the resolved position is mirrored onto it keeping the alignment (`top-start` → `bottom-start`), so the gap class follows.

Scoped to `top*` / `bottom*`. `left*` / `right*` sit beside their opener and keep their horizontal gap, so a vertical correction no longer restyles them onto a vertical side — that's covered by a test.

## Tests

`composables` 108 pass (2 updated, 2 added). The two updated ones are the LIB-2813 flip tests, which encoded the old `VIEWPORT_MARGIN` spacing; they now encode the margin-driven gap (zero in the harness, which mounts a bare element with no gap class) and additionally assert the reported side. New tests cover a non-zero gap class and the `left-start` no-restyle case.

Every `FzFloating` consumer's suite passes too: tooltip 69, dropdown 71, select 138, tab 79, typeahead 122, card-list 60, datepicker 130.

## Notes for reviewers

- The resolved position now reports the flipped side. That's the intended reading — a flipped panel *is* a panel on that side — but code comparing it against the requested `position` prop to detect "no flip happened" will now see them differ.
- A panel whose gap class is overridden away will touch its opener when it flips. That already happens when it's placed natively on that side; the flip simply no longer supplies a gap of its own.
- `nx run-many --target=test:unit` can't run in a worktree here (the `build` prerequisite fails for `icons`/`style`/etc. without `FONTAWESOME_PACKAGE_TOKEN`), so the suites above were run per-package through vitest directly.
- [LIB-2832](https://fiscozen.atlassian.net/browse/LIB-2832) is the companion: the `Tooltip Top Position` story asserts both the old placement *and* `mb-4`, so it needs updating on top of this. It is currently failing on main and blocking the pre-push hook.

[LIB-2831]: https://fiscozen.atlassian.net/browse/LIB-2831?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LIB-2813]: https://fiscozen.atlassian.net/browse/LIB-2813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LIB-2832]: https://fiscozen.atlassian.net/browse/LIB-2832?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/floating-margin-follows-flip.md
+++ b/.changeset/floating-margin-follows-flip.md
@@ -1,0 +1,18 @@
+---
+"@fiscozen/composables": patch
+---
+
+FzFloating: make a flipped panel's gap and gap class follow the side it actually landed on
+
+The opener-aware vertical correction shipped in 1.1.2 moves a panel to the other side of its opener when it does not fit. What it did not move was the spacing: the gap class is derived from the resolved position, so a `top` panel flipped below its opener kept `mb-4`, leaving the margin on the edge facing away from the opener.
+
+That class is load-bearing, not decorative. The position calculators take the gap from the element's own margin — `bottom*` leaves `y` on the opener's edge and lets `margin-top` shift the fixed element down, `top*` subtracts `margin-bottom` outright — while the correction spaced a flipped panel by `VIEWPORT_MARGIN` instead. So a flipped panel sat 8px from its opener where a natively-placed one on that side sits 16px, and carried a margin that did nothing on a fixed element positioned by `top`.
+
+Both halves now come from the same place:
+
+- the flip is spaced by the gap the requested side declares in its margin class, so a flipped panel is spaced exactly like a native one on that side — above the opener the gap is in the arithmetic, below it `y` sits on the opener's edge and the flipped side's `margin-top` supplies it;
+- the correction reports the side it landed on, and the resolved position is mirrored onto it (keeping the alignment: `top-start` becomes `bottom-start`), so the gap class follows.
+
+Only `top*` / `bottom*` placements are affected. `left*` / `right*` sit beside their opener and keep their horizontal gap, so a vertical correction no longer restyles them onto a vertical side.
+
+Two notes for consumers. The resolved position now reports the flipped side, which is the intended reading — a flipped panel _is_ a panel on that side — but code that compared it against the requested `position` prop to detect "no flip happened" will see them differ. And a panel whose gap class has been overridden away now touches its opener when it flips, exactly as it already did when placed natively on that side; the flip no longer supplies a gap of its own.
--- a/packages/composables/src/__tests__/useFloating.spec.ts
+++ b/packages/composables/src/__tests__/useFloating.spec.ts
@@ -390,11 +390,14 @@ describe('useFloating', () => {
       // The invariant is unchanged: the whole menu stays on screen.
       expect(floating.float.position.y).toBeGreaterThanOrEqual(VIEWPORT_MARGIN)
       expect(floating.float.position.y + ELEMENT_HEIGHT).toBeLessThanOrEqual(600 - VIEWPORT_MARGIN)
-      // It gets there by flipping to the other side of the opener (bottom edge one margin
-      // above opener.top = 540 - 8 - 100) rather than sliding up across it, which would
-      // have left the menu covering the control that opened it.
-      expect(floating.float.position.y).toBe(540 - VIEWPORT_MARGIN - ELEMENT_HEIGHT)
+      // It gets there by flipping to the other side of the opener rather than sliding up
+      // across it, which would have left the menu covering the control that opened it.
+      // The gap comes from the element's own margin (zero here, since the harness mounts a
+      // bare element with no gap class), so the menu's bottom lands on opener.top = 540.
+      expect(floating.float.position.y).toBe(540 - ELEMENT_HEIGHT)
       expect(floating.float.position.y + ELEMENT_HEIGHT).toBeLessThanOrEqual(540)
+      // The resolved position follows the flip, so the consumer's gap class does too.
+      expect(floating.actualPosition?.value).toBe('top-start')
     })
 
     it('flips a menu below its opener when it would overflow the top', async () => {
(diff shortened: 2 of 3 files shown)


packages/composables/src/composables/useFloating.ts
```diff
@@ -631,15 +674,34 @@ export const useFloating = (
...
 
       // Step 9: Apply boundary corrections
+      //
+      // Only vertical placements get the flip treatment: `left*` / `right*` sit beside the
+      // opener and keep their horizontal gap, so moving one vertically must not restyle it.
+      const vertical = isVerticalPosition(actualPosition.value!)
+      // The design gap lives in the margin class of the requested side, so read it from
+      // there rather than hard-coding 16px: `top*` carries it as `margin-bottom`,
+      // `bottom*` as `margin-top`.
+      const openerGap = actualPosition.value!.startsWith('top') ? margins.bottom : margins.top
```
francesca-vago, line 684: **This reads the gap from the requested side, but by the second pass the class already belongs to the flipped side — so the gap collapses to 0.**

Three facts combine:

1. `actualPosition` is reset from the prop at the top of every run (line 618), exactly as the comment above the `mirrorVertically` call says.
2. This line therefore always asks the *requested* side: for a `bottom-start` request it reads `margins.top`.
3. But after the first flip `FzFloating` has already swapped the class from `mt-4` to `mb-4`, because `getMarginClass` derives it from `actualPosition` (`FzFloating.vue:196-208`). So on every subsequent run `margins.top` is `0`, `aboveOpener` becomes `opener.top - 0 - element.height`, and the panel's bottom edge lands exactly on `opener.top`.

The second pass is not hypothetical: the `isOpen` watcher calls `setupEventListeners()` — which does `resizeObserver.observe(content)`, whose initial notification is delivered in the same frame — *before* calling `floating.setPosition()`, and there is a capture-phase scroll listener on top of that.

Measured (390x600, opener 540..580, 100px element, `bottom-start`, margin swapped between passes as the component would):

```
main    PASS 1  y=432  gap=8   →  PASS 2  y=432  gap=8
branch  PASS 1  y=424  gap=16  →  PASS 2  y=440  gap=0
```

For a vertical placement exactly one of the two margins is ever non-zero (`mt-4` for `bottom*`, `mb-4` for `top*`), so the gap does not need to know which side the class currently reflects:

```suggestion
      const openerGap = Math.max(margins.top, margins.bottom)
```

With that, both passes give 16px and all 108 tests stay green.

francesca-vago (commented):
Not blocking this, but the first note below wants to land before it merges: as it stands the change makes a flipped panel's gap **smaller** than what is on `main` today, not larger.

I reproduced the sequence `FzFloating` actually drives — pass 1, then the gap class following the flip, then a second `setPosition()` — against both `main` and this branch:

```
main    PASS 1  y=432  gap=8   resolved=bottom-start
        PASS 2  y=432  gap=8   resolved=bottom-start     <- stable
branch  PASS 1  y=424  gap=16  resolved=top-start        <- the intent
        PASS 2  y=440  gap=0   resolved=top-start        <- the steady state
```

Harness: 390x600 viewport, opener at 540..580, 100px element, requested `bottom-start`, with `margin-top: 16px` on pass 1 and `margin-bottom: 16px` on pass 2 — which is what `getMarginClass` produces once `actualPosition` has become `top-start`. Mechanism and a one-line fix are inline on the `openerGap` line; I verified the fix gives 16px on both passes with all 108 tests still green.

Everything else checks out as described: `applyBoundaryCorrections` reporting `flippedTo`, the `left*` / `right*` scoping, `mirrorVertically` keeping the alignment, and the suite at 108.

### A side effect worth putting in the changeset

`FzSelect` and `FzTypeahead` derive the panel's max-height from `actualPosition` — `pos.includes("bottom") ? calc(100vh - bottom - buffer) : ${top}px` (`FzSelect.vue:462`, `FzTypeahead.vue:379`). Reporting the flipped side means a panel that flips above its opener is now capped by the space *above* instead of the space below, which is a latent bug fixed rather than one introduced. But it is a behaviour change in two components that neither the changeset nor the tests mention — "every consumer's suite passes" is not the same as "no consumer's behaviour changed", and select's 138 tests do not exercise the flipped case.

### Diff hygiene

Roughly 150 lines of this diff are pure prettier reformatting — quote style, trailing whitespace, wrapping — sitting on top of a behavioural change in the repo's hub package. It makes the review harder and `git blame` noisier. I would split it into its own commit.

### Merge order

If the `Tooltip Top Position` story is already red on `main`, LIB-2832 wants to land with this or before it, otherwise `main` stays red and everyone's pre-push hook stays blocked. That story is also the only automated check that would notice the class following the flip: it asserts both `mb-4` and the placement (`Tooltip.stories.ts:485-488`). I have not verified the red-on-main claim — I did not run the Storybook suite.

packages/composables/src/__tests__/useFloating.spec.ts
```diff
@@ -423,13 +426,84 @@ describe('useFloating', () => {
...
+        opener: { domRef: ref(mockOpener) }
+      })
+
+      const floating = useFloating(args)
+      await floating.setPosition()
+      await nextTick()
+
+      // Flipped above: the 16px gap is in the arithmetic, because a `margin-bottom` cannot
+      // move a fixed element positioned by `top`.
+      expect(floating.float.position.y).toBe(540 - 16 - ELEMENT_HEIGHT)
```
francesca-vago, line 469: Every flip test calls `setPosition()` once, and the defect I left on `useFloating.ts:684` lives in the second pass — so the suite cannot currently see it. Worth extending this test rather than adding a new one: swap the margin onto the flipped side, which is what the component does once `actualPosition` becomes `top-start`, and assert the gap survives.

```ts
      expect(floating.float.position.y).toBe(540 - 16 - ELEMENT_HEIGHT)
      expect(floating.actualPosition?.value).toBe('top-start')

      // FzFloating now re-renders with the flipped side's class: mt-4 -> mb-4. Any
      // reposition after that — the content ResizeObserver's initial delivery, a scroll,
      // a resize — must not lose the gap.
      mockElement.style.marginTop = ''
      mockElement.style.marginBottom = '16px'
      await floating.setPosition()
      await nextTick()
      expect(floating.float.position.y).toBe(540 - 16 - ELEMENT_HEIGHT)

      mockElement.style.marginBottom = ''
```

On the current branch that second assertion fails with `expected 440 to be 424`.

.changeset/floating-margin-follows-flip.md
```diff
@@ -0,0 +1,18 @@
+---
+"@fiscozen/composables": patch
```
francesca-vago, line 2: `@fiscozen/composables` is at 1.1.3, and `actualPosition` is public API whose meaning changes here. Your own notes for consumers spell it out — code comparing it against the requested `position` will now see them differ — and `FzSelect` / `FzTypeahead` switch max-height branch because of it. `minor` looks like the more honest bump for anyone pinned to `~1.1.x`.

PR #434, opened 2026-08-07
test(storybook): LIB-2860 stop asserting a tooltip position that cannot happen

francesca-vago:
Jira: [LIB-2860](https://fiscozen.atlassian.net/browse/LIB-2860)

`Tooltip Top Position` fails on `main` — `expected 124 to be less than or equal to 45` — which makes the Storybook suite red and blocks everyone's pre-push. The engine is right and the test is wrong.

## What is actually happening

The story puts its trigger ~45px from the top of the page, so there is no room above it for the tooltip. Since LIB-2813 (#425) and LIB-2825 (#427) the engine places the panel where it fits rather than off-screen, so a `position="top"` tooltip in that layout lands below the trigger — and the assertion demands it be above.

Proof it is the layout and not the engine: add 300px of padding above the trigger, touch nothing else, and the test passes.

**Why it used to pass:** because of the very bug LIB-2825 removed. With the collision bounds collapsed by `body`'s rect, the clamp pinned the panel against the top margin, and `bottom ≈ 48` slipped under the `45 + 5` threshold. The assertion was satisfied by a panel detached from its trigger — it was encoding the defect.

## The change

Two stories instead of one, and no engine code touched:

- `TooltipTopPosition` gets room above it. The padding is load-bearing, and the comment now says so, so nobody "cleans it up" later.
- `TooltipTopWithoutRoom` covers the cramped case and asserts what matters: the panel stays on screen **and** stays attached to its trigger.

That second assertion is the point. My first version only checked "fully on screen", which the old pinned-to-the-top behaviour would also have satisfied — it would not have locked the fix in. Being adjacent to the trigger is what actually changed.

## Verification

- Tooltip stories: 11/11 green.
- Full pre-push suite: 745 tests green, 66 files (it was red before this change).
- Only story files change, so no changeset is due.

## Note for whoever merged the engine fixes

I could not work out how this reached `main`: the pre-push hook runs the whole suite and the failure is deterministic here, across clean `main`, a branch merging it, and retries. Either the storybook gate was skipped through its documented escape hatch, or the geometry differs on another machine. Worth a look — if it is the second case, the old assertion was environment-sensitive in a way the new one is not.

[LIB-2860]: https://fiscozen.atlassian.net/browse/LIB-2860?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/storybook/src/stories/overlay/Tooltip.stories.ts
+++ b/apps/storybook/src/stories/overlay/Tooltip.stories.ts
@@ -416,6 +416,7 @@ export const TooltipStylingRegression: Story = {
   }
 }
 
+
 /**
  * Tests tooltip positioning with bottom position and verifies margin class
  */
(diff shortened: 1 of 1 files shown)


francesca-vago:
Duplicata di #432, che risolve lo stesso rosso — e lo risolve meglio: la sua story del caso stretto verifica anche che la classe del gap segua il lato dove il pannello è atterrato, coprendo LIB-2813 e LIB-2831 insieme, mentre questa si fermava a "resta attaccato al trigger".

Colpa mia: ho trovato `Tooltip Top Position` rosso mentre riallineavo #433 su main e sono partito a sistemarlo senza controllare le PR aperte, dove #432 stava già da prima. Chiudo questa; il branch resta in piedi, cancellatelo pure da qui.

Niente da riportare su #432.

PR #450, opened 2026-09-07
feat(thumbnail): LIB-2918 FzThumbnail — an image in a box the caller sizes

Trebh:
Jira: [LIB-2918](https://fiscozen.atlassian.net/browse/LIB-2918)

New package `@fiscozen/thumbnail`. The design system had no way to show an arbitrary image, and
`` is in `forbiddenTags` at every atomic-design layer — so there was neither a component to
compose nor a legal fallback.

Only three components render an `` today, and none is usable as a standalone image:
`FzAvatar` (round, hardwired `size-32`/`size-44`, built for people) and
`FzRadioCard`/`FzCheckboxCard`, where the image is an illustration inside a selectable card, via
`imageUrl` + `imageAlt` + `object-cover h-full w-full`. **This extracts and generalises that
second pattern rather than inventing anything.**

It is a divergence already in production: in the frontoffice customer chat, image attachments
render as file cards, indistinguishable from a document — and not distinguishable
iconographically either, because the icon kit has no image glyph. Only the filename tells the
user it is a photo.

## The API

```vue
<FzThumbnail src="/receipt.jpg" alt="Scontrino di marzo" width="158px" height="108px" />
```

| Prop | Type | Default | |
|---|---|---|---|
| `src` | `string` | — | Image URL. |
| `alt` | `string` | — | Accessible name. **Required.** |
| `width` / `height` | `string` | — | Any CSS length. |
| `aspectRatio` | `string` | — | e.g. `'16 / 9'`, when only one dimension is known. |
| `radius` | `'none' \| 'sm' \| 'base' \| 'lg' \| 'xl'` | `'base'` | `base` is 4px, the design's value. |
| `bordered` | `boolean` | `false` | 1px `grey-100` border. |
| `scrim` | `boolean` | `false` | Translucent veil over the image only. |
| `overlayPosition` | `'top-start' \| 'top-end' \| 'bottom-start' \| 'bottom-end' \| 'center'` | `'bottom-end'` | Where the `overlay` slot sits, 8px in. |
| `placeholderIcon` | `string` | `'file'` | Shown on load error. |
| `loading` | `'lazy' \| 'eager'` | `'lazy'` | Native hint. |

Plus an `error` event carrying the failing `src`, and an `overlay` slot.

## Two decisions worth a reviewer's eye

**The dimensions are props, not a class.** The card's constraint is that the caller imposes the
size and the component hardwires none, and `width`/`height`/`aspectRatio` satisfy that — a unit
test asserts nothing on the root sets a dimension and that the style attribute is absent by
default. But the reason they are *props* is that the consuming apps forbid `class` outright at
the organism, template and page layers, and `style` anywhere above an atom. A class-only sizing
contract would make this component unusable exactly where a page needs it, so deciding it inside
the component is the only way the capability exists for those layers. That is the same argument
LIB-2951 made for `FzNavbar`'s `elevation`, and it applies to `overlayPosition` for the same
reason: without it, a page could not place the download button either.

It also means the Storybook stories are class-free, as a DS story should be.

**The scrim is package CSS, not `bg-grey-500/20`** — and that is not a style preference. **A
Tailwind opacity modifier on a design-system colour token generates no rule at all.** The colours
go through `filterTokensByType`, which wraps them in `var()` for runtime theming, so `grey.500`
is the string `var(--grey-500, #596167)`; Tailwind cannot decompose a `var()` into channels and
drops the candidate instead of emitting it — silently, with no warning and no build error.
Verified with the Tailwind CLI against the DS preset:

```
input:  bg-core-black/40  bg-grey-500/20  bg-black/40  bg-core-black
output: .bg-black\/40   { background-color: rgb(0 0 0 / 0.4) }   ← Tailwind's own palette, hex
        .bg-core-black  { background-color: var(--core-black, #2c282f) }
        (no .bg-core-black\/40, no .bg-grey-500\/20)
```

So the scrim is declared in the package's `<style>`, from the token's value, with
`--fz-thumbnail-scrim` as the consumer's escape hatch — the `FzDialog` backdrop precedent. The
`<style>` block and the MDX both explain why, so nobody "simplifies" it back into a utility.

## Found on the way: a shipped bug in `@fiscozen/layout`

`FzAppTemplate.vue:265` and `FzSidebarTemplate.vue:187` both draw their mobile drawer backdrop
with `bg-core-black/40`, neither is compensated by package CSS, and `bg-core-black\/40` appears
in no built CSS file — **both backdrops render fully transparent.** Click-to-dismiss works; the
dimming and the modal affordance do not exist. These are the only two sites in the repo applying
an opacity modifier to a token colour. Filed as
**[LIB-2995](https://fiscozen.atlassian.net/browse/LIB-2995)** and deliberately **not** fixed
here — different package, different card.

## Design reference

The prototype is `AttachmentImage.vue` in `tool.fiscozen.agentic-design`, branch
`design/equip-cliente` (the card points at the wrong repo). Its geometry re-expressed on DS
tokens: `rounded-[4px]` → `rounded` (`--rounded-base`, 4px, exact), `border-fz-grey-100` →
`border-grey-100` (`#e9edf0`, exact), and the scrim `rgba(74,85,101,0.2)` → `grey-500` at 20%,
which is the nearest DS colour since `#4A5565` is not one.

**On the card's "take the measurements from Figma":** that requirement lands on the consumer, not
here. Since this component hardwires no size, 158×108 versus the later 168px-tall
column-filling variant is a decision for `FzpChatAttachmentCard` in `@fzp`. The only
Figma-derived values this package needs are the radius, the border and the scrim, resolved above.

The spec also revealed a second use case the card does not mention — annotation 39, *"Preview
immagini"*, a fixed-size preview in the composer with an **"x"** to remove it. That is why the
overlay is a free-form slot with a positionable corner rather than a `downloadable` boolean.

## Checks

- **45 unit tests**, **100%** statements / branches / functions / lines on the component
- **9 Storybook play functions**, measured in Chromium rather than asserted on class lists: the
  scrim's computed `rgba(89, 97, 103, 0.2)`, the resolved 4px radius, the overlay layer's
  `pointer-events: none` with its child still clickable, `aspectRatio` resolving 320px → 180px,
  and the placeholder box still measuring exactly 158×108 *after* a load failure — a broken URL
  leaving no hole is the point, so it is measured, not assumed
- The `Cropping` story puts a light, a dark, a very wide and a very tall image in the same box,
  as the DoD asks, and asserts all four keep it
- Repo-wide: **44 projects'** unit tests, **48 projects'** build (skip-cache, 0 TS errors),
  **773** Storybook tests
- Changeset present — a `minor` on a new package at 0.1.0, so it first publishes as 0.2.0, the
  shape `@fiscozen/popover` used
- MDX docs at `apps/storybook/src/FzThumbnail.mdx`, package README, and the story under
  `Media/FzThumbnail`

## Verified rather than assumed

- **The icon kit still has no image glyph**, re-checked against **1.0.430** (the version
  LIB-2983 just landed, not the 1.0.421 the card's evidence came from): 156 names in
  `classic/solid`, 157 in `classic/regular`, zero matches for image/photo/camera/picture. The 18
  new names over 1.0.421 are auth/user glyphs (`browser`, `id-badge`, `unlock`,
  `right-to-bracket`, …), nothing image-related, nothing removed. So `placeholderIcon` defaulting
  to `file` stands. There is **no `download` glyph either** — the docs and stories use
  `arrow-down-to-line`.
- **`FzIconButton`'s `tooltip` prop is deprecated**, so every example here uses `ariaLabel` for
  the accessible name.
- `@fiscozen/icons` is a **peer** dependency and is in `rollupOptions.external`, matching
  `card-list`.

## Deliberately not in scope

`FzRadioCard` and `FzCheckboxCard` keep their inline `<picture>`. Refactoring them onto
`FzThumbnail` would be genuinely DRY, but it adds a dependency edge from radio + checkbox to a
brand-new package, a cascade bump, and visual-regression risk on two shipped components — for a
card that only asks for the new component. Worth its own card if wanted.

Unblocks **LIB-2912** (`FzpChatAttachmentCard`), whose README documents the file-card fallback as
temporary and says nothing else about that component has to change. It stays blocked on this
package being *published*, not on it existing.

[LIB-2918]: https://fiscozen.atlassian.net/browse/LIB-2918?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LIB-2995]: https://fiscozen.atlassian.net/browse/LIB-2995?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


francesca-vago:
## Review

I re-ran the PR's own claims rather than reading them, and they hold up.

**The Tailwind claim is true.** Local Tailwind CLI (3.4.19) against the DS preset, `@tailwind utilities` in, the four candidates from the description plus the classes this component actually uses:

```
in:  bg-core-black/40  bg-grey-500/20  bg-black/40  bg-core-black  border-1  p-8  rounded  size-full
out: .size-full  .rounded  .border-1  .bg-black\/40  .bg-core-black  .p-8
     → no .bg-core-black\/40, no .bg-grey-500\/20
```

**The `@fiscozen/layout` bug is real and confined.** `FzAppTemplate.vue:265` and `FzSidebarTemplate.vue:187` are the only two sites in the repo applying an opacity modifier to a token colour. Right call leaving them to LIB-2995.

**Every token resolves:** `spacing-8` = 8px, `border-1` (19 pre-existing uses), `rounded` = 4px, `grey-100/300/500`, `core-white`. `rgba(89, 97, 103, .2)` is exactly `#596167` at 20%. And no Storybook wiring is needed — the tailwind `content` glob already wildcards `./node_modules/@fiscozen/**/src/**`, `media/` exists, `consultant.jpg` is in `public/`.

---

### 1. The docs contradict the behaviour — blocking

`README.md:21`, `FzThumbnail.mdx:52` and the SFC docstring all say:

> With no dimension at all the box collapses to zero and nothing is visible.

I replicated the markup with the Tailwind preflight and measured it in a 600px container:

```
box: 600x450    img: 600x450    (natural 400x300)
```

`height: 100%` against a parent whose height is `auto` is an indefinite percentage, so it resolves to `auto` and the image renders full-width at its natural ratio. It is **fully visible**. That is a perfectly sensible default being described as broken in three files, and no test covers the zero-prop case — the sizing test asserts the `style` attribute is absent, which is a different thing.

Either fix the three claims and document what actually happens, or lock the intended behaviour in with a test. As written it is a declaration nothing honours.

### 2. "A load error does not leave a hole" silently depends on an explicit `height` — blocking

Same container, no dimension props:

| | box |
|---|---|
| image loads | 600×450 |
| image fails | **600×17** |

The placeholder has no intrinsic size, so it collapses exactly where the image did not. `LoadError` measures 158×108 only because the story passes them. So the guarantee holds for the chat feed and silently does not hold for a caller who relies on the natural-ratio behaviour from #1 — and the asymmetry is documented nowhere. It is the one case the placeholder exists to prevent, so it is worth either a minimum height on the placeholder or an explicit line in the docs plus a test.

### 3. The sources do not pass the project's Prettier

`prettier --check` with `packages/prettier-config/prettier.config.js` (`semi: false`, `singleQuote: true`, `printWidth: 100`):

```
[warn] FzThumbnail.vue
[warn] index.ts
[warn] types.ts
```

All three use double quotes, semicolons and 80 columns. Nothing gates this — the pre-push hook runs changeset + build + storybook only — so it merges fine, but the first `pnpm --filter @fiscozen/thumbnail format` rewrites all three files wholesale. Worth running now while the diff is still this PR's.

### 4. The scrim freezes the light theme

`rgba(89, 97, 103, 0.2)` bakes in `--grey-500`'s fallback, so a consumer overriding the token gets every other grey following and the scrim not.

The precedent cited is the worse of the two in `FzDialog`: next to `.fz-dialog__backdrop` (hardcoded rgba) sits `dialog::backdrop { background: var(--core-black, …); opacity: 0.8 }`, which keeps theming. The scrim is an empty overlay div, so `opacity` has no downside here:

```css
.fz-thumbnail__scrim {
  background: var(--fz-thumbnail-scrim, var(--grey-500, #596167));
  opacity: 0.2;
}
```

Same rendered result, one less frozen value, and the `<style>` block's reasoning stays valid — it is still not `bg-grey-500/20`, for exactly the reason it documents.

### 5. `imgProps` declares keys it ignores

The type is the full `ImgHTMLAttributes`, so `imgProps: { src: '…' }` type-checks and does nothing — there is even a test asserting the silent no-op. `Omit<ImgHTMLAttributes, 'src' | 'alt' | 'loading'>` turns that into a compile error, which is what the prop's own docstring promises.

Related: `onError` is **not** overridden. Vue's `mergeProps` chains handlers, so an `onError` in `imgProps` fires alongside the internal one. The docstring says the error handler wins over anything named there; it does not.

### Minor

The stories pass `onDownload` / `onRemove` / `onError` as args on a component that declares none of them, so they land as DOM listeners on the root box and show up in the autodocs args table. `border-solid` on the bordered box is redundant — the preflight already sets `border: 0 solid` on `*`.

---

The rest is solid, and several decisions are better than what they replace: required `alt` with an explicit `alt=""`, the accessible name outliving the image, `failed` resetting on a new `src`, the inert overlay with pointer events handed back to its children, and `imgProps` as the way past a single root element. The `Cropping` story sampling the decoded bitmap instead of trusting geometry plus `object-fit` is the strongest test in the set — that is the bug class a cropping story exists to catch.

#1 and #2 are the two I would want fixed before merge; #3 and #5 are a minute each.

🤖 Review assisted by [Claude Code](https://claude.com/claude-code)

Trebh:
Thanks — all five reproduce. Fixed in 4f09978, and I re-measured each one first rather than taking the write-up on trust.

### #1 and #2 — confirmed, and now measured in the repo

Your numbers replicate exactly. Chromium, 600px container, 400×300 image:

| unsized box | |
|---|---|
| image loads | 600×450 |
| image fails | **600×20** |

The three "collapses to zero" claims now describe the natural-ratio behaviour, and "a broken URL does not leave a hole" is stated for what it actually is — a guarantee about a *sized* box, with the reason the asymmetry exists (an image has an intrinsic ratio to fill an auto height with; a placeholder does not).

Both are now locked at the tier that can see them. A new `Unsized` story renders one good `src` and one broken one, unsized, and asserts against the parent's `clientWidth` rather than a number so the story stays class-free — it reproduces your finding at 1280×960 in the repo's own Chromium. I mutation-checked it (`300/400` → `301/400` → `expected 960 to be 963`), so the measurement is real rather than vacuously true. The unit test alongside it locks the classes those measurements follow from, so adding a min-height has to be a deliberate change to both.

**No min-height, though** — that would hand the component a size of its own, which is the one thing it refuses. Documented and tested instead.

### #4 — took the variant you suggested, plus one property

```css
.fz-thumbnail__scrim {
  background-color: var(--fz-thumbnail-scrim, var(--grey-500, #596167));
  opacity: var(--fz-thumbnail-scrim-opacity, 0.2);
}
```

`--grey-500: #596167` really is defined in `packages/style/output/global.css:230`, so the override you describe is a live capability the baked `rgba()` was throwing away. The extra property is because folding the 20% into `opacity` alone would quietly change what `--fz-thumbnail-scrim` *means*: a consumer setting `rgba(0,0,0,0.6)` would get 0.12 effective, and could never reach opaque. Package is unpublished, so there is no compat cost to having both. The `<style>` block's `bg-grey-500/20` reasoning is untouched — that was never the part in question. Story now asserts `rgb(89, 97, 103)` + `opacity: 0.2` separately.

### #5 — both halves, and `onError` is the interesting one

Type is now `Omit<ImgHTMLAttributes, 'src' | 'alt' | 'loading' | 'onError'>`.

`onError` is in that list rather than merely documented because you're right that it does not lose — I proved it before believing it:

```
imgProps: { onError: spy } → spy called 1 time
```

`mergeProps` collects `onXxx` into an array, so the smuggled handler fires *alongside* the internal one. Nothing breaks (the placeholder still appears), but "the component's handler wins" was a lie, and excluding the key is what makes it moot. There is now a test asserting the chaining explicitly, so the reason for the exclusion is recorded rather than folded away.

One caveat on the compile-time half: `packages/thumbnail/tsconfig.json` excludes `src/__tests__`, so `vue-tsc` never sees the spec and a `@ts-expect-error` there would be dead weight. The type is the guard at consumer call sites; the spec keeps asserting the runtime behaviour behind it.

### #3 — did it, but the stated consequence doesn't hold

All four files reformatted (you missed `__tests__/FzThumbnail.spec.ts`, which was in the same state).

But `pnpm --filter @fiscozen/thumbnail format` would **not** have rewritten them, and nothing else would either:

```
$ prettier --find-config-path packages/thumbnail/src/types.ts
[error] Can not find configure file
```

`prettierrc.js` is not a filename Prettier discovers — it looks for `.prettierrc*` or `prettier.config.*`. So `format` runs on Prettier *defaults*, which is exactly what those files already matched; that is why they looked "wrong" in the first place.

**Correcting myself on the scope, though** — I said above that one package carries `prettier.config.js` and the rest carry the undiscoverable name. That was wrong: the one `prettier.config.js` is `packages/prettier-config/prettier.config.js`, the shared config itself, not a package-level one. The 46 packages actually split three ways, and I only happened to probe `thumbnail` and `dialog`, one from each broken camp:

| camp | | drift: `packages/*/src` files failing the shared config |
|---|---|---|
| **A** | 20 packages with `.prettierrc.js` — **discovered**, house style applies | 68 / 154 — **44%** |
| **B** | 16 packages with `prettierrc.js` — **not discovered** | 142 / 150 — **95%** |
| **C** | 10 packages with no config file at all | 78 / 78 — **100%** |

288 of 382 in total. So the honest version of my claim is narrower than "nothing is configured" but worse in substance: **camp B's 95% against camp A's 44% is causal evidence, not correlation.** Where the config resolves, the code mostly stays in style even with nothing enforcing it; where it is silently ignored, `pnpm format` doesn't merely fail to help — it actively rewrites the file *away* from the documented style, and 95% of camp B has been rewritten that way. `@vue/eslint-config-prettier/skip-formatting` takes ESLint out of it, and the hooks only run tests, build, changeset and Storybook, so nothing anywhere notices.

I still made the change here, on the narrower ground that the PR's own stories are single-quoted and semicolon-free and CLAUDE.md says the same, so the PR was internally inconsistent with itself.

**Filed as [LIB-3010](https://fiscozen.atlassian.net/browse/LIB-3010)** (Bug, Backlog) rather than riding along here. All 36 per-package configs turn out to be byte-identical bare spreads of `@fiscozen/prettier-config` with zero real overrides, and there is no root config — so the fix is one root `prettier.config.js`, delete all 36, realign the plop template so camp B cannot repopulate, one mechanical reformat commit into the existing `.git-blame-ignore-revs`, and a `format:check` so the drift is visible next time. Without that last part it is 288 files fixed and back to square one.

### Minors

Split the `fn()` spies out of the `v-bind` in the three render functions and disabled them in the args table — you're right that they were landing on the root box as DOM listeners. (`onError` stayed: it's a declared emit.)

**Declining `border-solid`**, though. It is not redundant noise by this repo's standard — 30 uses across `FzButton`, `FzCard`, `FzDivider`, `FzCheckbox`, `FzCheckboxCard`, `FzDatepicker` and more. Dropping it here would make the new package the odd one out, which is a worse outcome than one belt-and-braces declaration.

### Checks

52 unit tests, 100% statements/branches/functions/lines on the component. 776 Storybook tests. 44 projects' unit tests. 48 projects' build, 0 TS errors. Storybook builds, so the MDX changes compile.

🤖 Reply assisted by [Claude Code](https://claude.com/claude-code)



[LIB-3010]: https://fiscozen.atlassian.net/browse/LIB-3010?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

francesca-vago (approved):
Tutti e cinque i rilievi verificati e chiusi.

